PR #84, opened 2025-11-07
Obol-agent template

bussyjd:


Padraic-O-Mhuiris (approved):

cmd/obol/main.go
```diff
@@ -151,6 +155,40 @@ GLOBAL OPTIONS:
...
 			// ============================================================
+			// Obol Agent Commands
+			// ============================================================
+			{
+				Name:  "agent",
+				Usage: "Manage Obol Agent",
+				Subcommands: []*cli.Command{
+					{
+						Name:  "init",
+						Usage: "Initialize Obol Agent with Google API key",
```
OisinKyne:
```suggestion
						Usage: "Initialize the Obol Agent by fetching an API key",
```

internal/agent/agent.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	applySecret := exec.CommandWithOutput(kubectlPath, "--kubeconfig", kubeconfigPath, "apply", "-f", "-")
+	applySecret.SetStdin(strings.NewReader(string(secretYAML)))
+	if err := applySecret.Run(); err != nil {
+		return fmt.Errorf("failed to create Google API key secret: %w", err)
+	}
+
+	l.Success("Google API key secret created")
+	l.Success("Obol Agent initialized successfully")
+	l.Info("The Obol Agent deployment will now have access to Google API services")
```
OisinKyne:
```suggestion
	l.Success("Obol Agent initialized. Open the Agent interface on the Stack UI to begin leveraging your Obol Agent. https://obol.stack#agent")
```

Skip the URL until something like that works. We can also consider launching the browser window with xdg-open at some stage (not now)

internal/embed/defaults/obol-stack-defaults/templates/obol-agent.yaml
```diff
@@ -0,0 +1,283 @@
...
+              cpu: 500m      # 0.5 CPU cores
+              memory: 1Gi    # 1 GiB RAM
+            limits:    # Maximum allowed resources
+              cpu: 2000m     # 2 CPU cores
+              memory: 4Gi    # 4 GiB RAM
+
+---
+#------------------------------------------------------------------------------
+# Service - Exposes the Obol Agent within the cluster
+# Access the agent at: http://obol-agent.default.svc.cluster.local:8000
```
OisinKyne:
```suggestion
# Access the agent at: http://obol-agent.agent.svc.cluster.local:8000
```

internal/embed/defaults/obol-stack-defaults/templates/obol-agent.yaml
```diff
@@ -0,0 +1,283 @@
...
+              path: /health
+              port: http
+            initialDelaySeconds: 10
+            periodSeconds: 5
+
+          # Resource limits prevent the agent from consuming too many cluster resources
+          resources:
+            requests:  # Minimum guaranteed resources
+              cpu: 500m      # 0.5 CPU cores
+              memory: 1Gi    # 1 GiB RAM
```
OisinKyne: i think in theory not specifying a min allows the stack to be as light as possible. Do we want to reserve a full gb of ram for an agent people might not use all the time? if it doesn't break stuff i'd remove

OisinKyne (approved):
Mostly LGTM except lets remove google from the customer facing interface at a minimum if not all the internals. I stopped making suggestions after a while when the idea is communicated. Suggestion for @agaskrobot around allowing the Agent component to auto-show if e.g. an anchor is in the url like `obol.stack#agent`.


I made a ticket https://github.com/ObolNetwork/obol-stack/issues/117  for us to make this command work with stdin as well, for users who need to be guided through it. TBH, it might end up with us doing an oauth flow for the users between google and our CLI command.

PR #85, opened 2025-11-07
Initialise release workflow

Padraic-O-Mhuiris:


.github/workflows/release.yml
```diff
@@ -0,0 +1,174 @@
...
+          - goos: linux
+            goarch: arm64
+          - goos: darwin
+            goarch: amd64
+          - goos: darwin
+            goarch: arm64
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
OisinKyne: v5 exists now

PR #103, opened 2025-11-17
Networks refactor

Padraic-O-Mhuiris:


bussyjd (changes requested):
### 🛑 Issues

1.  **Logging Framework Removal**:
    -   Stripped out `internal/logging` and `internal/executor` and replaced all structured logs with `fmt.Println/Printf`.
    -   **Why this is bad**: We lose colored output, log levels (Success/Info/Error), and most importantly, **file-based logging** for debugging. Users won't have logs to share when things go wrong.
    -   **Request**: Please revert the removal of the logging framework or integrate it into the new network system.

2.  **Language**:
    -   `plan.md`: `TODO Fucking aztec name override`
    -   **Request**: Please remove this immediately.

3.  **Architecture Discrepancy (Ephemeral vs Persistent)**:
    -   The `plan.md` and comments in `internal/network/network.go` state:
        > Installed networks: $OBOL_CONFIG_DIR/networks/<network>/helmfile.yaml
    -   However, `Install()` in `internal/network/network.go` does this:
        ```go
        tmpDir, err := os.MkdirTemp(...)
        // ...
        defer os.RemoveAll(tmpDir)
        ```
    -   **Impact**: The network is installed from a temporary directory that is **deleted immediately after the command finishes**. This means:
        -   We lose the record of what was installed.
        -   `Delete()` (which is currently a TODO) will have nothing to delete from `$OBOL_CONFIG_DIR`.
        -   Users cannot manually inspect or modify the installed helmfile.

4.  **Incomplete Implementation**:
    -   `internal/network/network.go`: `Delete()` is just a TODO printing to stdout.
    -   If we merge this, users can install networks but never clean them up properly.

### 🔍 Code Quality

-   **`internal/network/parser.go`**: The regex-based parsing of Helm templates (`regexp.MustCompile`) is clever but brittle. Ensure it handles whitespace variations robustly.
-   **`cmd/obol/network.go`**: The dynamic flag generation relies on the `PARSE_PREFIX` convention (e.g. `AZTEC_`). Make sure this is documented clearly for anyone adding new networks.

Please address the critical issues, especially the architecture mismatch and the logging regression.

PR #110, opened 2025-11-24
Networks refactor

Padraic-O-Mhuiris:


bussyjd:
- EnvVar struct naming is a bit confusing..maybe `TemplateField`
- Either enforce single-word network names or always require the slash separator and make the network name parsing a helper (used twice).

- extractTemplateFields() needs to return deterministic sorting
- We don't have `network list --installed` or similar?
- Missing Validation on Template Execution - If a user provides a malformed value that breaks YAML syntax, the error occurs later during helmfile sync rather than at install time:
```
  if err := tmpl.Execute(&buf, templateData); err != nil {
      return fmt.Errorf("failed to execute values template: %w", err)
  }
  // No YAML validation of buf.Bytes() before writing
  ```
  - In values.yaml.gotmpl:
```
  @default Annotation Ambiguity - In values.yaml.gotmpl:

  # @default
  # @description Deployment identifier...
  id: {{.Id}}

```
  The @default annotation with no value means "has a default" but the actual default comes from petname generation in Go code.
  
  - Needs user confirmation for network confirmation:
eg:  "Are you sure you want to delete ethereum-prod?"

- Do we want to keep the `helmfile.yaml.gotmpl` in the deployment directory after install?

PR #118, opened 2025-12-09
Created app install/sync/list/delete flow

Padraic-O-Mhuiris:
Installs *.tgz chart url's to local applications directory as a local chart referenced by helmfile with auto-generated `values.yaml`. Syncs via helmfile sync as expected. Other facilitating commands included. Uses petname namespace strategy to maintain configuration-deployment correspondence.


OisinKyne:
also, i cant figure out how to manage the obol-frontend app with this, plausibly because its a default but we maybe do need an upgrade path for it too

PR #139, opened 2026-02-09
fix(llm): route OpenClaw through LLMSpy gateway

bussyjd:
## Summary

- Fix LLMSpy deployment: the init container and volume config now correctly copy both `llms.json` and `providers.json` (root cause of the crash / scale-to-0)
- Update default model name from `glm4` to `glm-4.7-flash` in the LLMSpy ConfigMap
- Route OpenClaw's Ollama provider through LLMSpy (`llmspy.llm.svc:8000`) instead of directly to Ollama (`ollama.llm.svc:11434`)
- Wire `agentModel` into the Helm chart's generated `openclaw.json`

## Test plan

- [x] LLMSpy pod starts and stays Running (1/1 ready, readiness probe passes)
- [x] Both `llms.json` and `providers.json` present in `/home/llms/.llms/`
- [x] Direct POST to LLMSpy `/v1/chat/completions` returns valid OpenAI-format JSON from Ollama
- [x] OpenClaw config (`openclaw.json`) shows `baseUrl` pointing to LLMSpy
- [x] Ollama logs confirm requests arriving from LLMSpy pod IP (full proxy chain verified)

### DIFF
--- a/internal/embed/embed.go
+++ b/internal/embed/embed.go
@@ -21,8 +21,10 @@ var infrastructureFS embed.FS
 //go:embed all:networks
 var networksFS embed.FS
 
-// CopyDefaults recursively copies all embedded infrastructure manifests to the destination directory
-func CopyDefaults(destDir string) error {
+// CopyDefaults recursively copies all embedded infrastructure manifests to the destination directory.
+// The replacements map is applied to every file: each key (e.g. "{{OLLAMA_HOST}}") is replaced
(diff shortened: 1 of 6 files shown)


OisinKyne:
Is it just that you had ollama on your host previously? or is the a reason it can't go in the stack? imo without good reason i see this as a con for the efficacy of the stack if it has to be on host. I'd much rather we put it in the stack and as an easy opt in for like `obol agent init local` or something.

PR #195, opened 2026-02-19
Erpc skill

agaskrobot:
Add local-erpc skill for interacting with the obol-stack eRPC gateway

Adds a new Claude Code skill documenting how to talk to the local eRPC instance deployed with every obol-stack cluster.

Covered:

- URL patterns for all access contexts (host via obol.stack, in-cluster, port-forward)
- Common JSON-RPC call examples (eth_blockNumber, eth_getBlockByNumber, eth_call, eth_getLogs, etc.)
- Per-request directives (X-ERPC-Skip-Cache-Read, X-ERPC-Use-Upstream, etc.)
- Health check and Prometheus metrics access via port-forward
- eRPC status/debugging commands (get pods, logs, describe)
- Routing architecture diagram (host → Traefik → eRPC → upstream)
- Working Python query script (scripts/example.py) — stdlib only, supports --network, --method, --params, --skip-cache, --port-forward


OisinKyne:
merged my ethereum-networks skill for now, we can iterate soon

PR #267, opened 2026-03-17
Delete duplicate openclaw. stop tunnelling ui to public internet. mak…

OisinKyne:
…e sale UI. needs frontend pr too.

I also have a PR for frontend, and will open another pr with just a plan for preferring litellm models (i have more than one in ollama and no ability to tell it to prefer the one i want to use. 

I think selling inference is in the wrong place in the UI imo. Something on the agent card should be for selling the agent, not the upstream inference. We also need to be able to specify what litellm upstream. I'd maybe sell my local inference but not my anthropic api key inference.


OisinKyne:
Replaced

PR #325, opened 2026-04-08
fix: allow obolup.sh to install openclaw via Docker when Node.js is missing

bussyjd:
## Summary

`obolup.sh` hard-blocked on missing Node.js even when Docker was available. The Docker fallback in `install_openclaw()` was never reached because `check_prerequisites()` exited first.

- Prerequisites: only fail if BOTH npm AND docker are missing
- `install_openclaw()`: try npm first, fall back to `docker create` + `docker cp` extraction

## Context

Hit this on DGX Spark nodes (Ubuntu ARM64 with Docker but no Node.js). Bootstrap failed at the prerequisite gate despite Docker being available.

## Test plan
- [x] `bash -n obolup.sh` syntax check passes
- [x] Manually tested: skips Node.js check when Docker available

### DIFF
--- a/obolup.sh
+++ b/obolup.sh
@@ -110,7 +110,9 @@ command_exists() {
 check_prerequisites() {
 	local missing=()
 
-	# Node.js 22+ / npm — required for openclaw CLI (unless already installed)
+	# Node.js 22+ / npm — preferred for openclaw CLI install.
+	# If missing, install_openclaw() will fall back to Docker image extraction.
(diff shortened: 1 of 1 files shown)


bussyjd:
Superseded by the validated integration branch \ and the \ prerelease cut from it. The release-candidate branch now carries the tested sell → discover → buy → settle path, updated docs/skills, and the final x402/buy-side fixes.

PR #347, opened 2026-04-17
feat(model): add master token accessor

bussyjd:
## Summary

Adds `obol model token` as a direct accessor for the LiteLLM master token stored in-cluster.

## Why

Production-readiness testing found that API consumers currently have to discover the LiteLLM master token by reading the `litellm-secrets` Secret manually via `obol kubectl get secret ...`. This PR turns the existing internal accessor into a supported CLI path.

## What changed

- added `model token` subcommand
- prints the LiteLLM master token in human mode
- returns `{"token": ...}` in JSON mode
- added command-structure coverage in `cmd/obol/model_test.go`

## Validation

- `go test ./cmd/obol`

### DIFF
--- a/cmd/obol/model.go
+++ b/cmd/obol/model.go
@@ -23,10 +23,33 @@ func modelCommand(cfg *config.Config) *cli.Command {
 		Commands: []*cli.Command{
 			modelSetupCommand(cfg),
 			modelStatusCommand(cfg),
+			modelTokenCommand(cfg),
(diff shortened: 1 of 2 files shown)


OisinKyne (approved):

OisinKyne:
Don't love re-using token for the third time (token the crypto, token the llm gas unit, token the auth), but we'll live for now

PR #339, opened 2026-04-11
feat: add exact OBOL Permit2 buy-side path

bussyjd:
## Summary

Adds the exact-path OBOL payment flow on top of the existing x402 buy-side stack.

This version includes:
- conflict resolution against current `main`
- the token UX and registry changes from [PR #340](https://github.com/ObolNetwork/obol-stack/pull/340)
- the exact-path OBOL gas/storage report

Scope of this PR:
- seller-side token selection with `--token [USDC|OBOL]`
- exact EVM OBOL path using `permit2` + `eip2612GasSponsoring`
- buyer-side generic signed-payment handling instead of ERC-3009-only auth reconstruction
- local fork-token integration coverage on a Base Sepolia Anvil fork
- exact-path sizing and gas analysis

This PR intentionally keeps packs / subscriptions / sessions as an upgrade path, not part of the implementation surface.

## UX changes from PR 340

This branch now includes the PR 340 direction:
- `--obol-token` has been replaced by `--token [USDC|OBOL]`
- token selection resolves through a whitelist registry in `internal/x402/tokens.go`
- default token remains `USDC`
- `OBOL` resolves to Permit2 metadata on Ethereum mainnet, with env override support for local fork deployments

## Why exact first

`exact` is still the lowest-risk path to shipping OBOL commerce in the current stack:
- aligned with the existing x402 verifier/facilitator flow
- preserves the current trust-minimized settlement model
- works with OBOL today via `permit2` + `eip2612GasSponsoring`
- avoids introducing long-lived session / escrow state into the first release

### Seller path
- add `--token [USDC|OBOL]`
- emit canonical token metadata into ServiceOffer payment terms
- keep default USDC behavior unchanged

### Buyer path
- store/replay full signed x402 payloads, not only legacy ERC-3009 auth fields
- support OBOL `permit2` payloads with `eip2612GasSponsoring`
- hard-cap Permit2 auth pools to `500` under the current ConfigMap-backed storage model
- improve user-facing output so OBOL amounts are displayed as OBOL, not USDC

### Local validation
- deploy a fork-local OBOL-compatible ERC20Permit token on Anvil
- validate sell -> buy -> sidecar -> settle end-to-end without any public OBOL deployment
- drain a 3-request exact pack and log all settlement receipts

## Exact-path gas report

Measured on a forked Base Sepolia Anvil with a fork-local OBOL token.

### 3-request exact pack benchmark

Per-settlement receipts:
1. `gasUsed=136993`, `gasWei=598512279518`
[...]

PR #355, opened 2026-04-19
First commit towards a demo sell command

OisinKyne:
Target v0.9.0, needs further testing and refinement (but committing for visibility)

Problem to be solved: 0 -> selling something takes too much knowledge for someone just checking the stack out

Proposed solution: some demo commands to make it easy to sell something basic

### DIFF
--- a/.github/workflows/docker-publish-storefront.yml
+++ b/.github/workflows/docker-publish-storefront.yml
@@ -1,0 +1,95 @@
+name: Build and Publish Public Storefront
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+    paths:
+      - 'web/public-storefront/**'
+      - 'Dockerfile.public-storefront'
+      - '.github/workflows/docker-publish-storefront.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: storefront-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract image metadata
+        id: meta
+        uses: docker/metadata-action@902fa8ec7d6ecbf8d84d538b9b233a880e428804 # v5.7.0
+        with:
+          images: ${{ env.REGISTRY }}/obolnetwork/obol-stack-public-storefront
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=sha,prefix=
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
+          labels: |
+            org.opencontainers.image.title=obol-stack-public-storefront
+            org.opencontainers.image.description=Public tunnel storefront for Obol Stack
+            org.opencontainers.image.vendor=Obol
+            org.opencontainers.image.source=https://github.com/ObolNetwork/obol-stack
+
+      - name: Build and push
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
+        with:
+          context: .
+          file: Dockerfile.public-storefront
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha,scope=storefront
+          cache-to: type=gha,scope=storefront,mode=max
+          provenance: true
+          sbom: true
+
+  security-scan:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+
+    steps:
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
+        with:
+          image-ref: ${{ env.REGISTRY }}/obolnetwork/obol-stack-public-storefront:${{ github.sha }}
+          format: 'sarif'
+          output: 'trivy-results.sarif'
+          severity: 'CRITICAL,HIGH'
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@b13d724d35ff0a814e21683638ed68ed34cf53d1 # main
+        with:
+          sarif_file: 'trivy-results.sarif'
+        if: always()
--- a/.github/workflows/docker-publish-x402.yml
+++ b/.github/workflows/docker-publish-x402.yml
@@ -10,12 +10,15 @@ on:
       - 'internal/x402/**'
       - 'internal/serviceoffercontroller/**'
       - 'internal/monetizeapi/**'
+      - 'internal/demo/**'
       - 'cmd/x402-verifier/**'
       - 'cmd/x402-buyer/**'
       - 'cmd/serviceoffer-controller/**'
+      - 'cmd/demo-server/**'
       - 'Dockerfile.x402-verifier'
       - 'Dockerfile.x402-buyer'
       - 'Dockerfile.serviceoffer-controller'
+      - 'Dockerfile.demo-server'
       - 'go.mod'
       - 'go.sum'
       - '.github/workflows/docker-publish-x402.yml'
@@ -53,6 +56,10 @@ jobs:
             image: obolnetwork/serviceoffer-controller
             dockerfile: Dockerfile.serviceoffer-controller
             description: ServiceOffer reconciler for Obol Stack monetization
+          - component: demo-server
+            image: obolnetwork/demo-server
+            dockerfile: Dockerfile.demo-server
+            description: Demo HTTP services for Obol Stack sell demo
 
     steps:
       - name: Checkout
@@ -124,6 +131,8 @@ jobs:
             image: obolnetwork/x402-buyer
           - component: serviceoffer-controller
             image: obolnetwork/serviceoffer-controller
+          - component: demo-server
+            image: obolnetwork/demo-server
 
     steps:
       - name: Run Trivy vulnerability scanner
--- a/.gitignore
+++ b/.gitignore
@@ -27,11 +27,15 @@ build/
 .workspace/
 .workspace-*/
 
-# Go binary
+# Go binaries
 /obol
+/demo-server
+/serviceoffer-controller
+/x402-verifier
+/x402-buyer
 
 # Dependencies
-/node_modules/
+node_modules/
 /vendor/
 
 # Environment variables
--- a/Dockerfile.demo-server
+++ b/Dockerfile.demo-server
@@ -1,0 +1,10 @@
+FROM golang:1.25-alpine AS builder
+WORKDIR /src
+COPY go.mod go.sum ./
+RUN go mod download
+COPY . .
+RUN CGO_ENABLED=0 go build -o /demo-server ./cmd/demo-server
+
+FROM gcr.io/distroless/static-debian12:nonroot
+COPY --from=builder /demo-server /demo-server
+ENTRYPOINT ["/demo-server"]
--- a/Dockerfile.public-storefront
+++ b/Dockerfile.public-storefront
@@ -1,0 +1,27 @@
+FROM node:22-alpine AS base
+
+FROM base AS deps
+WORKDIR /app
+COPY web/public-storefront/package.json web/public-storefront/package-lock.json ./
+RUN npm ci
+
+FROM base AS builder
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY web/public-storefront/ .
+RUN mkdir -p public && npm run build
+
+FROM base AS runner
+WORKDIR /app
+ENV NODE_ENV=production
+RUN addgroup --system --gid 1001 nodejs && \
+    adduser --system --uid 1001 nextjs
+
+COPY --from=builder /app/public ./public
+COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
+COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
+
+USER nextjs
+EXPOSE 3000
+ENV PORT=3000
+CMD ["node", "server.js"]
--- a/cmd/demo-server/main.go
+++ b/cmd/demo-server/main.go
@@ -1,0 +1,78 @@
+// Command demo-server runs a lightweight HTTP server for obol sell demo.
+//
+// The demo type is selected by the DEMO_TYPE environment variable:
+//   - hello:  proof-of-payment echo (no external dependencies)
+//   - blocks: basic chain data from eRPC
+//   - oracle: chain analysis with gas statistics from eRPC
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/ObolNetwork/obol-stack/internal/demo"
+)
+
+func main() {
+	demoType := envOr("DEMO_TYPE", "hello")
+	port := envOr("PORT", "8080")
+	erpcURL := envOr("ERPC_URL", "http://erpc.erpc.svc.cluster.local/rpc/base")
+
+	var handler http.HandlerFunc
+	switch demoType {
+	case "hello":
+		handler = demo.HelloHandler()
+	case "blocks":
+		handler = demo.BlocksHandler(erpcURL)
+	case "oracle":
+		handler = demo.OracleHandler(erpcURL)
+	default:
+		log.Fatalf("unknown DEMO_TYPE: %q (expected hello, blocks, or oracle)", demoType)
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET /", handler)
+	mux.HandleFunc("GET /health", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprintln(w, "ok")
+	})
+
+	addr := net.JoinHostPort("", port)
+	srv := &http.Server{
+		Addr:              addr,
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+
+	log.Printf("demo-server type=%s listening on %s", demoType, addr)
+
+	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("listen: %v", err)
+		}
+	}()
+
+	<-ctx.Done()
+	log.Println("shutting down...")
+
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	_ = srv.Shutdown(shutdownCtx)
+}
+
+func envOr(key, fallback string) string {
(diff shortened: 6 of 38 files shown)


bussyjd:
Pushed `5ad56d7` with the local fixes for the two demo blockers I hit during review:

- demo Deployments now use locally imported images instead of pulling `ghcr.io/obolnetwork/demo-server:latest`
- `blocks` and `oracle` now get an eRPC URL with the required `/rpc/<network>` path
- the public storefront image now builds successfully in `OBOL_DEVELOPMENT`
- the storefront Deployment also uses the locally imported image

Retest on this PR branch after cleanup:

- `go test -v ./cmd/obol`
- `go test -v ./internal/demo ./internal/serviceoffercontroller`
- clean `obol stack up`
- `obol sell demo hello --name review-hello --chain base-sepolia ...`
- `obol sell demo blocks --name review-blocks --chain base-sepolia ...`
- `obol sell demo oracle --name review-oracle --chain base-sepolia ...`

Results:

- all three demo backends deploy successfully now
- all three public `/services/...` routes return `402`
- direct backend calls for `hello`, `blocks`, and `oracle` all returned real JSON payloads
- the storefront root now serves successfully

Remaining issue:

- multiple demo services still compete over the singleton `/.well-known/agent-registration.json` path
- in my retest, `review-hello` became `Ready=True`, but `review-blocks` and `review-oracle` ended up `Registered=False` with `SingletonConflict`
- as a result, only the first demo showed up in `/api/services.json`

So the image/eRPC/storefront blockers are fixed on-branch, but I still think the multi-demo registration model needs another pass if the goal is to have several demos coexist on the storefront.

bussyjd:
Follow-up from local retest after layering the shared ERC-8004 registration fix from #363 on top of this PR branch.

Original findings still stand for this PR as written:

1. `demo-server:latest` needs `imagePullPolicy: IfNotPresent` or the locally imported image is ignored and the demo pods fail with `ErrImagePull/ImagePullBackOff`.
2. `blocks` / `oracle` need an eRPC URL with `/rpc/<network>` instead of the root service URL.

I applied those fixes locally together with the shared-registration changes from #363 and reran the demos on the PR355 branch.

### Local retest result with both sets of changes applied

Commands:
- `obol sell demo hello --name review-hello --chain base-sepolia ...`
- `obol sell demo blocks --name review-blocks --chain base-sepolia ...`
- `obol sell demo oracle --name review-oracle --chain base-sepolia ...`

Observed behavior after convergence:
- all three demo backends deployed successfully
- all three `ServiceOffer`s reached `Registered=True`
- all three `ServiceOffer`s reached `Ready=True`
- all three public `/services/...` routes returned `402`
- `/.well-known/agent-registration.json` contained all three demo endpoints in one shared `services[]` array
- `/api/services.json` contained all three demo services

Representative status change on non-owner offers after the #363 logic was layered in:
- `review-blocks`: `Registered=True`, message `Shared registration via demo/review-hello is active`
- `review-oracle`: `Registered=True`, message `Shared registration via demo/review-hello is active`

So the current read is:
- this PR still needs the image/eRPC fixes called out above
- once those are in, the shared-registration work from #363 does resolve the singleton `/.well-known/agent-registration.json` conflict for the multi-demo flow

bussyjd:
## Added test coverage (`7a0ef17`)

Four focused test additions. All placed in files/packages **not** touched by the other open PRs (#339 Permit2, #364 buy-autorefill, #353 cluster-down, #360 flow-11, #356 trivy), so there's no expected merge conflict.

### Coverage delta

| Package                              | Before  | After      |
|--------------------------------------|--------:|-----------:|
| `internal/demo`                      | 85.0%   | **94.6%**  |
| `internal/serviceoffercontroller`    | 18.4%   | 18.8%      |

The serviceoffercontroller number barely moved because most of that package is reconciler loops that need fake-client integration tests — the new branches lit up are the only statements that were reachable from pure unit tests.

---

### 1. `internal/demo/oracle_helpers_test.go` — pure math the oracle demo sells

The oracle demo's whole value is chain-analysis math, yet six helpers had zero direct tests before. A refactor that swaps `pct > 90` for `pct >= 90` would have silently mislabeled a congested block as "busy" for paying customers.

```
  utilizationLabel tiers (now locked in)
  ──────────────────────────────────────
      0 ──┐
          │ "low"
     40 ──┤    ← boundary tested at 40.0 AND 40.0001
          │ "moderate"
     70 ──┤    ← boundary tested at 70.0 AND 70.0001
          │ "busy"
     90 ──┤    ← boundary tested at 90.0 AND 90.0001
          │ "congested"
    100 ──┘

  Also verified: weiToGwei precision (0, 1, 1e9, 1.5e9, 12.3457e9),
                 hexToUint64 (empty, 0x10, 10, 0xffff, 0xdeadbeef),
                 hexToBigInt (0x0, 0x10, ff, 0x3b9aca00),
                 trimQuotes (quoted, unquoted, "", `"`, empty),
                 safeDivFloat (incl. div-by-zero → 0),
                 computeGasStats (empty, single, multi-element, input-mutation).
```

### 2. `internal/demo/rpc_errors_test.go` — `rpcCall` error branches

Before this, only the success and transport-failure paths were covered. A JSON-RPC `error` field, a malformed body, or an HTTP 500 from eRPC all flowed through untested code.

```
  rpcCall outcomes              before  after
  ─────────────────             ────── ─────
  Post fails (unreachable)  ──→  YES    YES
  ReadAll fails             ──→  no     no    (hard to trigger deterministically)
  Unmarshal fails           ──→  no     YES   ← "decode response"
  rpcResp.Error != nil      ──→  no     YES   ← "rpc error <code>: <msg>"
  HTTP 500 + RPC err body   ──→  no     YES   ← (verifier passes through)
  success                   ──→  YES    YES
```

### 3. `internal/demo/handlers_extra_test.go` — OracleHandler early-return + payment fallback

The oracle handler aborts to an errors-only body when `eth_blockNumber` fails — critical, because otherwise it would try to compute blocks from `blockNum=0` and return nonsense. The existing tests all used a mock that never errored.

```
  OracleHandler timeline (new test verifies early-return branch)
  ─────────────────────────────────────────────────────────────

  chainId ── OK ─┐
                 │
  blockNumber ── FAIL ──► respond(errors only) ──► return
                          ↑                        ↑
                          │                        │
                     assert forbidden keys    assert no further
                     absent from response     RPC calls fire
                     (recentBlocks,           (mock server fails
                     gasAnalysis, txVolume)   the test if any
                                              unexpected method
                                              is called)
```

Also covers `extractPayment` default-status fallback (untested before — existing test always set the header), `extractPayment` header passthrough, and `firstNonEmpty` (first-wins, fall-through, all-empty, nil, single).

### 4. `render_test.go` — `buildServiceCatalogJSON` filter pipeline

Before: one happy-path + one empty test. The filter pipeline has four exclusion rules, none of which were directly verified — a regression on any of them exposes internal state on the public storefront.

```
  offers (mixed)
      │
      ▼
  ┌──────────────────────────────┐
  │ nil? DeletionTimestamp?      │    now tested — all 4 exclusions
  │ paused annotation?           │    exercised in a single table.
  │ Ready condition != "True"?   │
  └──────────────────────────────┘
      │
      ▼
  sort by Name                        now tested — 3-offer reverse-order
      │                                 input produces alphabetical output.
      ▼
  ┌──────────────────────────────┐
  │ per-offer mapping:           │
  │  Description fallback        │    now tested — empty Registration.
  │                              │    Description yields auto-generated
  │                              │    "x402 payment-gated <type> service".
  │  per-mtok → Price non-empty, │
  │             PriceRaw empty   │    now tested.
  │  baseURL trailing slash      │    now tested — no //services endpoints.
  └──────────────────────────────┘
```

---

### Test run

```
$ go test -race -cover ./internal/demo/ ./internal/serviceoffercontroller/
ok  github.com/ObolNetwork/obol-stack/internal/demo                  1.482s  coverage: 94.6%
ok  github.com/ObolNetwork/obol-stack/internal/serviceoffercontroller 1.896s  coverage: 18.8%
```

No changes to production code — tests only.

bussyjd:
## Public-release readiness: **NOT YET — 3 blockers**

I took the three-commit PR (`01bcd6a`, `7a0ef17`, `ada0dab`), rebuilt `obol`, `demo-server`, `obol-stack-public-storefront`, and `serviceoffer-controller` locally with `OBOL_DEVELOPMENT=true`, imported them into a running k3d cluster, and walked the "zero-to-selling" path end-to-end (hello, blocks, oracle; local + tunnel; storefront + skill catalog + services.json; delete cleanup).

The plumbing works — routing, 402 shape, handlers, storefront, JSON catalog, delete. But a first-time public user following the copy-paste flow **will not see a working demo**, for three reasons below.

---

### ✅ What I verified works

| Path | Evidence |
|---|---|
| `obol sell demo hello\|blocks\|oracle` deploys Namespace + Deployment + Service + ServiceOffer | all 3 reconcile `UpstreamHealthy=True`, `PaymentGateReady=True`, `RoutePublished=True` in ≤10 s |
| Unpaid `GET $TUNNEL/services/demo-<type>` returns HTTP **402** with valid x402 v2 pricing (`scheme=exact`, `network=eip155:84532`, USDC asset, `payTo`, `amount`) | `curl` at each endpoint |
| `blocks` handler concurrently fans out to eRPC → returns `blockNumber`, `gasPrice`, `chainId`, latest block body | direct port-forward to pod |
| `oracle` handler fetches last 5 blocks, computes min/max/avg gwei, gas utilisation %, tx volume | direct port-forward to pod |
| `/skill.md`, `/.well-known/agent-registration.json`, and the new `/api/services.json` are reachable on the tunnel hostname | 200 on each |
| Next.js storefront renders at tunnel `/` (replaces busybox) | 200 + rendered HTML |
| `obol sell delete demo-<type> -n demo --force` cleans Deployment + Service + ServiceOffer (demo-namespace branch) | confirmed empty after |
| Auto-starting the quick tunnel inside `sell demo` | prints URL, wires into `AGENT_BASE_URL` on the agent deploy |

### ❌ Blockers for public release

1. **Images don't exist on GHCR yet.** `ghcr.io/obolnetwork/demo-server:latest` and `ghcr.io/obolnetwork/obol-stack-public-storefront:latest` are only wired into the publish workflows (`docker-publish-x402.yml` matrix adds `demo-server`, and the new `docker-publish-storefront.yml`). Both workflows only run on `push` to `main`. A user who installs via `obolup.sh` today — i.e. **without** `OBOL_DEVELOPMENT=true` — will get `ImagePullBackOff` on both the demo pod and the tunnel storefront pod. After merge, wait for CI green on main and confirm GHCR has pullable `:latest` tags before announcing.

2. **Registration is forced on → service never reaches `Ready` → empty storefront.** `buildDemoServiceOffer` always sets `registration.enabled: true`. Against a stock `obol stack up`, the controller has no ERC-8004 signing key, the Registered condition sits at `AwaitingExternalRegistration` indefinitely, and therefore `isConditionTrue(status, "Ready")` never flips true. The immediate consequence, for the end-user trying the demo:
   - `/api/services.json` returns `[]`
   - `/skill.md` says *"No services currently available"*
   - the Next.js storefront shows an empty state
   - the 402 gate still works, so the CLI output *looks* correct — but the whole "public face" of the demo is blank.
   
   I re-ran with `spec.registration.enabled=false` patched in and everything flipped to Ready in ~2 s, storefront populated, `/api/services.json` returned the expected record. Either (a) default `enabled: false` for the demo command, (b) allow off-chain-only / `OffChainOnly` to count as Ready in `requestPhaseReady`, or (c) gate the ERC-8004 step behind an explicit `--register` flag on `sell demo`.

3. **Multiple demos can't coexist.** If a user runs `obol sell demo hello` then `obol sell demo blocks`, the second offer goes to `Registered=False / SingletonConflict`. Same root cause as (2): the stack enforces one registration owner. Also fixable by dropping `enabled: true` from the default demo spec.

### ⚠️ Polish worth fixing before release

- **`--chain` defaults to `base` (mainnet).** For a *demo* this is a footgun — the printed try-it example on chain `base` will cost a real buyer real USDC. Flip the default to `base-sepolia`, or at minimum gate `base` behind an explicit opt-in.
- **Quick-tunnel URL in `/.well-known/agent-registration.json`.** The quick tunnel hostname rotates every restart; if/when registration is re-enabled, every fresh `stack up` orphans the previous on-chain record. Document the requirement to use `obol tunnel login --hostname ...` before registering, or skip registration for quick-tunnel mode.
- **`hello` handler always returns `"payment":{"status":"paid"}`** regardless of whether the request came through the gate (I bypassed via port-forward and still got `paid`). For a *proof-of-payment* demo, derive the status from the presence of `X-Payment-Status: success` / `X-Payment-Tx` headers — otherwise the "echo what you got through" promise is a lie when inspected.
- **Try-it snippet uses `x402.client.x402_client`** which isn't a real import path in the published `x402` Python package. Replace with the actual API (`x402 httpx-client` wrapper or the curl-with-X-PAYMENT form), or point to the Coinbase x402 example.
- **`--namespace` override is missing on `obol sell demo`.** Every demo lands in `demo` ns; if a user wants to demo for two audiences they can't. Minor, but easy add given `sell http` already has the flag.
- **PR is 3 commits behind `main`** (the autorefill + ERC-8004 aggregation landed after this branch cut). Rebase before merging; the controller touches `reconcileRegistrationActive` / `selectRegistrationOwner` in both places.

### Minor

- `Dockerfile.demo-server` and `Dockerfile.public-storefront` both use `:latest` tags hard-coded in `sell.go` and `tunnel.go`. Consider pinning to commit SHAs once CI publishes, to make the demo reproducible across releases.
- `buildSkillCatalogConfigMap` now writes `services.json` + `.json:application/json` into `httpd.conf` — nice; worth a one-liner test that the MIME mapping actually fires (busybox httpd is picky about ordering).
- `servicesJSONRouteName = "obol-services-json-route"` is namespace-scoped to `x402` and **has no hostname restriction** — that's correct for tunnel access, but confirm the catalog deliberately exposes ready-offer metadata (name, payTo, chain) on the public internet. It does, and that's by design, but good to state explicitly in the PR description so reviewers don't flag it.

---

## Validated "human path" guide

The canonical zero-to-selling flow once the blockers above are addressed. I ran every step on a running `obol stack up` cluster (branch `pr-355-validation`, k3d `obol-stack-desired-seasnail`).

### 0. Pre-reqs (one-time)

```bash
# Fresh install (prod path — will need GHCR images to exist, see blocker 1)
curl -L https://obol.org/stack | bash

# Or, if you're running from source:
OBOL_DEVELOPMENT=true ./obolup.sh
```

Confirm the stack is up:

```bash
obol stack init && obol stack up        # ~2–3 min
obol kubectl get pods -A | grep -E "x402|traefik|cloudflared"
```

### 1. Deploy a demo

```bash
# Simplest — no chain deps, $0.00001/request
obol sell demo hello \
  --wallet 0xYourUSDCRecipient \
  --chain base-sepolia            # <-- recommend this as the default

# Live chain data — needs eRPC configured with the selected chain
obol sell demo blocks --wallet 0x... --chain base-sepolia

# Gas-stats oracle — same requirement as blocks
obol sell demo oracle --wallet 0x... --chain base-sepolia
```

The command prints: ServiceOffer created → tunnel URL → 4-part try-it block.

### 2. Verify the gate

```bash
export TUNNEL=$(obol tunnel status -o json | jq -r .url)

# Unpaid → 402 with pricing
curl -sS $TUNNEL/services/demo-hello | jq .
# -> { "x402Version": 2, "accepts": [ { "scheme":"exact", "network":"eip155:84532", ... } ] }
```

### 3. Verify the public face

```bash
# Storefront landing
open $TUNNEL/

# Machine-readable catalog (JSON)
curl -sS $TUNNEL/api/services.json | jq .

# Markdown catalog
curl -sS $TUNNEL/skill.md

# ERC-8004 registration doc
curl -sS $TUNNEL/.well-known/agent-registration.json | jq .
```

⚠ All three will show empty/0 services **until the ServiceOffer reaches `Ready`**. Today that means disabling registration (blocker 2):

```bash
obol kubectl patch serviceoffer demo-hello -n demo \
  --type merge -p '{"spec":{"registration":{"enabled":false}}}'

obol sell status demo-hello -n demo       # expect Ready=True
```

### 4. Pay for a request (buyer side)

```bash
# From a funded wallet on base-sepolia:
python - <<'EOF'
import httpx
# Use whatever x402 buyer library is current at release time
# (the README snippet "from x402.client import x402_client" in printDemoTryIt is not accurate
# against the published x402 Python package).
EOF
```

Or via the in-cluster `x402-buyer` sidecar + `buy.py` path (documented in CLAUDE.md).

### 5. Clean up

```bash
obol sell delete demo-hello -n demo --force
obol sell delete demo-blocks -n demo --force
obol sell delete demo-oracle -n demo --force
# Namespace is removed automatically when the last offer is deleted; tunnel stops.
```

---

## Recommendation

Don't ship this as part of a public v0.9.0 until at least the three blockers are closed. Specifically:

1. **Merge + wait for CI** so GHCR has `demo-server:latest` and `obol-stack-public-storefront:latest`.
2. **Default `registration.enabled: false`** on the demo ServiceOffer (or treat `OffChainOnly` as Ready for catalog purposes) — otherwise the storefront is permanently empty, which defeats the stated goal of the PR.
3. **Flip default chain to `base-sepolia`** for the demo subcommand.

With those three, this becomes a legitimately great 0-to-selling on-ramp. The code quality on the demo handlers, the JSON catalog, and the storefront integration is solid — tests cover the non-trivial branches well. Nice work @OisinKyne.

*— Validated against `pr-355` HEAD `ada0dab`, merge-base `899c605`, on k3d `obol-stack-desired-seasnail` with PR-built images loaded via `k3d image import`.*

bussyjd:
## Rebased onto current main (`f08651e`)

Worked from a fresh clone (kept the open worktrees untouched). All three commits cherry-picked, force-pushed.

### Conflict resolution (1 file)
`cmd/obol/sell_test.go` — kept both `TestIsTransientRegistrationError` (landed via #386) and the new `TestDemoRPCNetwork` + `TestBuildDemoResources_UsesImportedImageAndERPCPath` from this PR.

### Rebase fixups required by API drift on main
Folded into the original commits via `--autosquash`:

1. `cmd/obol/sell.go` — `openclaw.ResolveWalletAddress` swapped to `hermes.ResolveWalletAddress` (Hermes is the default agent runtime since #381; `openclaw` package is no longer imported in `sell.go`).
2. `internal/serviceoffercontroller/helpers_test.go` — removed `TestNewBigInt`, `TestContainsFinalizer`, `TestStatusFor_IdentityPassThrough`, and the five `TestLoadRegistrationSigningKey_*` tests. The functions they cover (`newBigInt`, `containsFinalizer`, `statusFor`, `loadRegistrationSigningKey`) were dead-code-trimmed in #375 / moved in #386.
3. `internal/serviceoffercontroller/purchase_pure_test.go` — dropped `TestHasStringInSlice` (helper removed in #375); renamed `TestNormalizePurchasedUpstreamURL` to `_EdgeCases` to coexist with the existing version that landed via #386 (this PR's table has wider edge-case coverage, so both kept); fixed type assertions on `auths[…]["signature"]` now that `preSignedAuthMaps` returns `[]map[string]any` (was `[]map[string]string`).
4. `internal/serviceoffercontroller/render_builders_test.go` — removed `TestBuildMiddleware`, `TestOwnerRef`, `TestOwnerRefFor_CustomAPIVersion`. The current architecture no longer renders a Traefik `Middleware` from the controller (HTTPRoute targets `x402-verifier` Service directly), and the typed `ownerRef` / `ownerRefFor` helpers were replaced by `ownerRefMap` / `ownerRefMapFor` returning `map[string]any`.

All other added tests (oracle math, RPC errors, catalog filters, render coverage) still apply and pass.

### Currency check (USDC vs OBOL Permit2)

Storefront UI is **token-agnostic** for display:
- `web/public-storefront/src/types.ts` — `Service` carries `price`, `network`, no token field; whatever string the backend ships is rendered verbatim.
- `PaymentFlow.tsx` — describes the x402 protocol generically (ERC-3009, X-PAYMENT). No USDC string anywhere.
- `ServiceCard.tsx` — renders `service.price`/`service.network` as plain strings.
- `page.tsx` — fetches `/api/services.json` (served by the controller's skill-catalog ConfigMap), no client-side currency assumptions.

**However, the JSON producer is USDC-only:**
- `internal/serviceoffercontroller/render.go::describeOfferPrice` (pre-existing in main, **not** introduced by this PR) hardcodes `"USDC/request"`, `"USDC/MTok"`, `"USDC/hour"` regardless of `offer.Spec.Payment.Asset.Symbol`. Since #355 now consumes this via `buildServiceCatalogJSON` and pipes the string straight to the storefront, an OBOL-Permit2-priced ServiceOffer would render as e.g. `"0.001 USDC/request"` on the public storefront even when the asset symbol is `OBOL`.
- `obol sell demo` writes a ServiceOffer **without** the new `payment.asset` block (no `Address`/`Symbol`/`TransferMethod`/`EIP712Name`). It only sets `scheme: exact`, `network`, `payTo`, `price.perRequest`. Demos therefore default to whatever the controller treats as the unspecified-asset baseline (USDC). That's fine for the proof-of-payment use case but means **the demo command doesn't exercise the OBOL Permit2 path**.

**Recommendation (follow-up, not blocking this PR):**
- Update `describeOfferPrice` to use `offer.Spec.Payment.Asset.Symbol` when set, falling back to `"USDC"`.
- Optionally add a `--token` flag to `obol sell demo` (mirroring `obol sell inference`) so demos can showcase OBOL Permit2.

Per the task constraints, I did not silently widen these — flagging only.

### Build + tests
- `go build ./...` — pass
- `go test ./... -timeout 180s` — all 25 packages pass, 0 failures
- `web/public-storefront`: `pnpm install` + `pnpm run build` — pass (Next.js 16.2.4, 3 static pages prerendered, TypeScript clean)

### Pushed
Force-pushed to `worktree-sell-demo`. PR will pick up the rebased tip `254f1a9`.

PR #357, opened 2026-04-20
fix(erc8004): send all SignTxRequest numeric fields as integers

HananINouman:
The Rust remote-signer deserializes nonce, gas_limit, max_fee_per_gas, max_priority_fee_per_gas, and value as u64 integers. Sending any of them as strings causes HTTP 422. The ChainID string→int64 fix in 9f53368 was correct but incomplete — all remaining numeric fields had the same problem.

Root cause: the bug was latent because the on-chain ERC-8004 registration path was only exercised when --register was explicitly passed. Once sell http made registration default-on (PR #345), every obol sell http call hit the remote-signer and failed with:
  nonce: invalid type: string "0", expected u64 at line 1 column 79

Fix: change Nonce/GasLimit/MaxFeePerGas/MaxPriorityFeePerGas/Value from string to uint64 in SignTxRequest, update RemoteTransactOpts to use .Uint64()/.Nonce()/.Gas() instead of fmt.Sprintf, update tests.

Verified: obol sell http with --register-name calls remote-signer successfully, Agent ID 5039 registered on-chain on base-sepolia.

### DIFF
--- a/internal/erc8004/signer.go
+++ b/internal/erc8004/signer.go
@@ -68,16 +68,16 @@ func (s *RemoteSigner) GetAddress(ctx context.Context) (common.Address, error) {
 }
 
 // SignTxRequest contains the fields for signing an EIP-1559 transaction.
-// chain_id is sent as a JSON integer (u64) to match the Rust remote-signer's
-// expected type — sending it as a string causes HTTP 422.
+// All numeric fields are sent as JSON integers (u64) to match the Rust
+// remote-signer's expected types — sending any of them as strings causes HTTP 422.
 type SignTxRequest struct {
 	ChainID              int64  `json:"chain_id"`
 	To                   string `json:"to"`
-	Nonce                string `json:"nonce"`
-	GasLimit             string `json:"gas_limit"`
-	MaxFeePerGas         string `json:"max_fee_per_gas"`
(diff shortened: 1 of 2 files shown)


bussyjd:
Pushed a follow-up test fix onto this branch to cover the regressed `RemoteTransactOpts` serialization path.

One non-blocking follow-up concern remains in `internal/erc8004/signer.go`: the new `.Uint64()` conversions for `Value`, `GasFeeCap`, and `GasTipCap` will silently truncate if a future tx exceeds `math.MaxUint64`. That does not affect the current ERC-8004 registration flow because the relevant `register(...)` / `setMetadata(...)` calls are `nonpayable`, so `value` should be zero here, but the helper is generic and would be safer with an explicit overflow check before conversion.

OisinKyne:
https://github.com/ObolNetwork/remote-signer/issues/5

Screenshot 2026-04-20 at 18 51 39

doing this in uint64 is legal json, but may mess up javascript setups. Considering openclaw is node.js and the frontend is javascript, should we go the other way, and have the remote signer parse strings to uint64?

bussyjd:
Follow-up is in #359. That PR aligns obol-stack with the canonical remote-signer string contract published in ObolNetwork/remote-signer#6, replacing the temporary integer-only request shape used here.

PR #381, opened 2026-04-24
Add Hermes default agent runtime

bussyjd:
## Summary
- Add Hermes as a first-class agent runtime with `obol hermes` commands while keeping manual `obol openclaw` support.
- Move stack-managed default agent paths (`stack up`, `agent init`, model sync, tunnel sync, x402 defaults) to the Hermes `obol-agent` singleton.
- Keep runtime state separated under `applications/hermes/<id>` and `applications/openclaw/<id>` with runtime-specific ports, namespaces, config, wallet metadata, and skills delivery.
- Harden dev stack testing for shared machines by deriving ingress URLs from generated k3d ports and excluding `.workspace*` from Docker build context.

## Validation
- `go test ./...`
- `bash -n flows/flow-01-prerequisites.sh flows/flow-02-stack-init-up.sh flows/flow-04-agent.sh flows/flow-07-sell-verify.sh flows/flow-08-buy.sh flows/flow-11-dual-stack.sh flows/lib.sh`
- `git diff --check`
- Shared-machine validation: prerequisites flow passed
- Shared-machine validation: stack init/up flow passed with occupied default ports and ephemeral ingress
- Shared-machine validation: agent flow passed, including Hermes token, health, chat completions, wallet, RBAC, and remote-signer checks

Note: the shared-machine test cluster was stopped after validation to reduce load.


OisinKyne:
Key questions for me:
- this time around, can we leave `obol hermes XXX` as the underlying binary, rather than the `obol openclaw cli` one? 
- they dropped a UI like today, should we support it?
- keen to try it tbh, and would pivot default sooner rather than later if it worked. 

We should sit down on a call with hanan, mae and others and maybe do some DREAMS.md / google design system stuff, and see if we can make the agent card more compelling than they currently are on our site. Particularly including what they're selling/buying.

bussyjd:
Follow-up from this branch after the frontend bump:

- Bumped the embedded frontend image to `obolnetwork/obol-stack-front-end:v0.1.17-rc.1`; `v0.1.16` is still the latest stable, but this RC contains the runtime-aware Hermes/OpenClaw agent UI.
- Tested on `spark2` against the shared port-collision setup. The frontend deployed with `v0.1.17-rc.1`, listed `hermes/obol-agent`, reported it healthy, and `/api/agents/chat` through the Obol frontend returned `hermes ui smoke ok` from Hermes.
- Confirmed `obol hermes skills ...` uses the Hermes-native binary path in the pod (`/opt/hermes/.venv/bin/hermes`), not the OpenClaw CLI.
- Grounded in `/Users/bussyjd/Development/R&D/hermes-agent`: Hermes has a separate `hermes dashboard` FastAPI/Vite app on port `9119` for config/API-key/session management, while the gateway API on `8642` is the OpenAI-compatible chat surface. I did not expose the native Hermes dashboard in this PR because Hermes explicitly treats public dashboard binding as sensitive; this PR keeps Obol's chatbox on the `:8642` OpenAI-compatible API and leaves native dashboard exposure as a product/security decision.
- Added a smoke-flow fix so k3d tests prefer the live Docker-published ingress port when an existing cluster keeps old port mappings after `stack init --force`.

Validation:
- Local `go test ./...` passed.
- Flow shell syntax and `git diff --check` passed.
- `spark2` `flow-04-agent.sh` passed all 17 steps after the live-port fix.
- PR checks are green.

bussyjd:
Added the native Hermes dashboard deeplink path:

- Hermes runtime now runs a `hermes-dashboard` sidecar using the Hermes-native `dashboard --host 0.0.0.0 --port 9119 --no-open --insecure` command, grounded against `/Users/bussyjd/Development/R&D/hermes-agent`.
- The Hermes service exposes both `http:8642` and `dashboard:9119`.
- The API/chat host stays `hermes-obol-agent.obol.stack`; the native UI gets a separate host: `hermes-obol-agent-ui.obol.stack`.
- Stack frontend is bumped to `obolnetwork/obol-stack-front-end:v0.1.17-rc.2`, which adds the Hermes dashboard deeplink.
- `flow-04-agent.sh` now validates the native UI deeplink by loading the Hermes dashboard root and checking for the injected `__HERMES_SESSION_TOKEN__` marker.

Validation:
- Local `go test ./...` passed.
- `bash -n` for smoke flows and `git diff --check` passed.
- Frontend PR #267 checks and `v0.1.17-rc.2` image/release checks are green.
- `spark2` deployed `v0.1.17-rc.2`, exposed `http:8642 dashboard:9119`, and `flow-04-agent.sh` passed 18/18 including `Hermes dashboard UI loaded: http://hermes-obol-agent-ui.obol.stack:38667`.
- PR checks are green.

cmd/obol/hermes.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/ObolNetwork/obol-stack/internal/hermes"
+	"github.com/urfave/cli/v3"
+)
+
+func hermesCommand(cfg *config.Config) *cli.Command {
+	return &cli.Command{
+		Name:    "hermes",
+		Aliases: []string{"herme"},
+		Usage:   "Manage Hermes agent instances",
+		Commands: []*cli.Command{
```
OisinKyne, line 17: I strongly think we should re-think here, and put most agent management commands under `obol agent onboard|auth(instead of token)|wallet|sync|delete|new|list` etc. And then we leave `obol hermes` as their CLI straight up. otherwise we'll have so much duplicated but slightly tweaked user facing code to maintain (`obol openclaw token` vs `obol hermes token` etc), and, no hermes/openclaw commands work out of the box by only prepending `obol `, they have to most likely add `obol ` and ` cli `, maybe `obol `, `<agent> cli`.

bussyjd: Agreed. I opened #384 to track the CLI UX consolidation explicitly.

Plan for the follow-up stacked PR: move Obol-owned lifecycle/auth/wallet/list/delete/sync/new commands under `obol agent`, and make `obol hermes` a native Hermes CLI passthrough against the selected deployed Hermes instance. Default target will be `obol-agent`; non-default Hermes instances can be selected explicitly. I’ll keep `obol openclaw` as the manual OpenClaw runtime path for now so the dual-runtime requirement remains intact.

bussyjd: Follow-up implementation is open as #385, stacked against this branch. It applies the #384 UX split: `obol agent` owns Obol lifecycle/auth/wallet management, and `obol hermes` becomes native Hermes CLI passthrough.

bussyjd:

PR #389, opened 2026-04-28
fix(model): unify LiteLLM model_name contract, remove double-strip

bussyjd:
## Summary

The flow-14 break that `ca820c9` band-aided was the visible symptom of a
deeper tangle: `obol model setup custom`, the LiteLLM `model_name`
convention, and `stripProviderPrefix` had no canonical contract. Three
copies of the helper across `internal/model`, `internal/hermes`, and
`internal/openclaw` each shaved a path component off model identifiers,
so a `custom/spark1-vllm/qwen36-fast` LiteLLM entry got walked all the
way down to `qwen36-fast` by the time the Hermes agent called
`/chat/completions`. LiteLLM only had the original key registered, so
every paid-model call surfaced as `400 "no healthy deployments for this
model"`.

This PR picks the contract explicitly, removes the redundant strip
layers, and adds tests that pin the round-trip property.

### Contract

> **LiteLLM `model_name` is the bare model identifier — the string the
> agent reads as `primary` and passes back as the `model` field on
> chat-completion calls. It must round-trip unchanged.**

This matches what every other code path in `buildModelEntries` already
does (Ollama, Anthropic, OpenAI explicit entries are all bare). The
band-aid commit (`ca820c9`) aligned custom endpoints with the same
convention; this PR makes the contract canonical and removes the
defensive stripping that was working around the old bug.

### What changed

- **`internal/model/model.go`** — extract `buildCustomEndpointEntry` so
  the entry shape is unit-testable. Document the contract on
  `AddCustomEndpoint` (bare `model_name`, `--name` is informational
  only). Drop the leftover `_ = name` bookkeeping from the band-aid.
- **`internal/model/rank.go`** — keep the unexported
  `stripProviderPrefix` for ranking heuristics (`localRank` still needs
  to parse `:9b` out of `anthropic/claude-…` shapes), but add a doc
  comment explicitly forbidding its use on round-trippable identifiers.
- **`internal/hermes/hermes.go`** — delete `stripProviderPrefix` and
  `stripProviderPrefixes` outright. `rankModels` now passes through to
  `model.Rank` without pre-stripping (which was the immediate
  double-strip bug). `configuredModels` returns the LiteLLM model list
  unchanged. The agent's `model.default` is now byte-identical to the
  LiteLLM ConfigMap entry.
- **`internal/openclaw/openclaw.go`** — already correct (delegates
  straight to `model.Rank`, only adds the `openai/` prefix on output for
  LiteLLM routing through the openai-compat slot). No changes needed —
  the openclaw `rankmodels_test.go` still passes unchanged, which is its
  own confirmation.
- **`cmd/obol/model.go`** — clarify the `--name` flag help to
  "informational only — LiteLLM keys the route by --model, not --name".
  No structural removal: the flag still surfaces in `obol model status`
  output and is accepted by existing flow scripts.

### Tests added

- `internal/model/rank_test.go::TestRank_PreservesProviderPrefixOnOutput`
  — pins the round-trip property at the `Rank()` boundary, including
  the legacy `custom/<name>/<model>` shape that broke flow-14.
- `internal/model/model_test.go::TestBuildCustomEndpointEntry` — covers
  the bare `model_name` + `openai/`-routing shape, the empty-key
  fallback (`api_key: none`), and that colon-tagged ids survive intact.
- `internal/hermes/rankmodels_test.go` — rewritten. The previous
  fixtures asserted the now-removed Hermes-side strip; the new ones
  assert the contract and add explicit guards for
  `custom/<name>/<model>` and provider-prefixed wildcards.
- `internal/hermes/hermes_test.go::TestGenerateConfig_PrimaryIsRoundTrippable`
  — end-to-end-ish: takes whatever LiteLLM might publish (bare Ollama
  tag, bare Claude id, bare OpenAI id, provider-prefixed wildcard
  expansion, custom endpoint bare) and confirms the agent's
  `model.default` is byte-identical.

### Lineage

- `ca820c9` — band-aid that dropped `custom/<name>/<model>` on the write
  side. Unblocked flow-14 but left the read-side double-strip latent.
- This PR — picks the contract, removes the redundant strips, locks the
  round-trip property in tests so neither side can drift again.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/model/... ./internal/hermes/... ./internal/openclaw/... ./cmd/obol/...` (all green)
- [x] Full repo `go test ./...` (all green)
- [ ] Manual: re-run `flow-14-live-obol-base-sepolia.sh` against a live
      `OBOL_LLM_ENDPOINT` to confirm `obol model setup custom` →
      `obol model sync` → agent chat works end-to-end
- [ ] Manual: confirm `obol model status` still shows the `--name`
      label for custom endpoints (label-only behavior preserved)

## Follow-ups (not in this PR)

- The `--name` flag is now strictly informational. We could either drop
  it (operators just supply `--model`) or repurpose it as a per-endpoint
  alias once LiteLLM gets multi-endpoint namespacing for the same model
  id. Tracking decision separately rather than coupling it to this fix.
- `internal/openclaw/SyncOverlayModels` is exported but has no caller in
  this branch's commit graph — likely dead code that drifted across the
  PR-377/381 merge. Worth confirming and removing in a separate cleanup.

### DIFF
--- a/cmd/obol/model.go
+++ b/cmd/obol/model.go
@@ -240,9 +240,9 @@ func modelSetupCustomCommand(cfg *config.Config) *cli.Command {
 		Name:  "custom",
 		Usage: "Add a custom OpenAI-compatible endpoint (validates before adding)",
 		Flags: []cli.Flag{
-			&cli.StringFlag{Name: "name", Usage: "Short name for the endpoint (e.g. my-vllm)", Required: true},
+			&cli.StringFlag{Name: "name", Usage: "Short label for the endpoint (informational only — LiteLLM keys the route by --model, not --name)", Required: true},
 			&cli.StringFlag{Name: "endpoint", Usage: "Full base URL (e.g. http://host:8000/v1)", Required: true},
-			&cli.StringFlag{Name: "model", Usage: "Model name at the endpoint", Required: true},
+			&cli.StringFlag{Name: "model", Usage: "Model identifier at the endpoint — this is also the LiteLLM model_name the agent will call", Required: true},
 			&cli.StringFlag{Name: "api-key", Usage: "API key (optional, some endpoints don't require it)"},
 			&cli.BoolFlag{Name: "no-sync", Usage: "Skip the agent model sync (batch with other model commands, then run `obol model sync` once)"},
 		},
--- a/internal/hermes/hermes.go
+++ b/internal/hermes/hermes.go
@@ -987,11 +987,20 @@ func syncObolSkills(cfg *config.Config, id string) error {
 	return nil
 }
 
+// configuredModels returns the agent-facing model list and the primary model
+// name. Both are returned as round-trippable LiteLLM `model_name` strings:
+// the agent passes `primary` back as the `model` field on chat-completion
+// calls, and LiteLLM matches by exact string against the entries in the
+// returned slice. NO provider-prefix stripping happens on this path —
+// LiteLLM `model_name` is the contract identifier end-to-end.
+//
+// See internal/model/model.go (AddCustomEndpoint, buildModelEntries) for
+// where the bare-name convention is written into the LiteLLM ConfigMap.
 func configuredModels(cfg *config.Config, u *ui.UI) ([]string, string, error) {
 	models, err := model.GetConfiguredModels(cfg)
 	if err == nil && len(models) > 0 {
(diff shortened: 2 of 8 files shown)


OisinKyne (approved):
changed base because the other base is the one it looks like you want to merge to main

PR #388, opened 2026-04-28
feat: flow-14 live Base Sepolia + network status reachability + flow-13 hardening

bussyjd:
## Summary

Picks up the v0.9.0-rc1 token-report follow-ups so we can run the OBOL Permit2 path against **live Base Sepolia** (not Anvil), and proactively surface the eRPC pin staleness that triggered PR #387's setMetadata revert.

Targets `integration/pr377-pr381` (the v0.9.0-rc1 branch) so the next RC bump can ship this as part of the same release line.

### 1. `obol network status` reachability probe (`feat(network)` commit)

`obol network status` now sends an `eth_chainId` JSON-RPC to every upstream in the eRPC config (parallel, 2s timeout) and warns when:
- An upstream is unreachable (the symptom of the report's "stale custom pin pointed at a dead Anvil fork" scenario)
- An upstream answers with a chain id that doesn't match the chain it's pinned to (re-pointed fork)

Output suggests the actionable recovery: `obol network remove <name>`. `--no-probe` opts out for callers in tight loops.

```
Reachability warnings:
  upstream "custom-84532-0" (chain 84532) at http://127.0.0.1:43117 is unreachable: connection refused

If any of these are stale custom pins from a previous test run, drop them with:
  obol network remove <chain-name>   # e.g. obol network remove base-sepolia
```

### 2. `flow-14-live-obol-base-sepolia.sh` (`feat(flow-14)` commit)

Live-network sibling of flow-13. Same dual-stack topology, but:

| | flow-13 (Anvil fork) | flow-14 (live) |
|---|---|---|
| Chain RPC | Anvil fork of 84532 | https://sepolia.base.org |
| Facilitator | local x402-rs | https://x402.gcp.obol.tech |
| OBOL token | `forge create ForkObolToken` per run | `OBOL_TOKEN_BASE_SEPOLIA` env (pre-deployed) |
| Bob funding | `mint()` from deployer | `transfer()` from `BOB_FUNDING_PRIVATE_KEY` |
| Registration | disabled | **enabled** (exercises PR #387 `WaitForAgent` on the OBOL path) |

Required env vars fail fast at the top of the script (no gas spent if missing):
- `REMOTE_SIGNER_PRIVATE_KEY` — Alice's seller key (Base Sepolia ETH holder)
- `OBOL_TOKEN_BASE_SEPOLIA` — deployed ERC20Permit address
- `BOB_FUNDING_PRIVATE_KEY` — buyer's wallet, must already hold real OBOL

[...]

PR #387, opened 2026-04-27
fix(erc8004): wait for read-side consistency before setMetadata

bussyjd:
## Summary

Fixes a `setMetadata` revert observed on **live Base Sepolia** during `obol sell register`:

```
[base-sepolia] (chain ID 84532)
    Registry: 0x8004A818BFB912233c491871b3d84c89A494BD9e
    Using direct on-chain registration via remote-signer...
    Wallet:   0x2FbFe6cF08Ac224f97915ecF07eE29Be0b213f51
    Agent ID: 5196
    Owner:    0x2FbFe6cF08Ac224f97915ecF07eE29Be0b213f51
  ! failed to set x402 metadata: erc8004: setMetadata tx: execution reverted
    CAIP-10:  eip155:84532:0x8004A818BFB912233c491871b3d84c89A494BD9e
  ✓ Agent registered on 1/1 networks.
```

## Diagnosis

| Check | Result |
|---|---|
| `cast nonce 0x2FbFe6cF... --rpc-url https://sepolia.base.org` (post-incident) | nonce advanced — only `register` broadcast |
| `cast call ownerOf(5196)` (or `getAgentWallet(5196)`) on the registry | returns `0x2FbFe6cF...` ✓ token IS minted |
| `cast call setMetadata(5196, "x402", '{"x402":true}') --from 0x2FbFe6cF` (today, live) | returns `0x` — would succeed |
| `cast call setMetadata(999999999, "x402", 0x01) --from 0x2FbFe6cF` | reverts with data `0x7e273289…` |
| `cast 4byte 0x7e273289` | **`ERC721NonexistentToken(uint256)`** |

So the colleague's diagnosis is correct: the revert came from **`eth_estimateGas` simulation, not on-chain**. The simulation hit a state where token 5196 didn't yet exist.

```
        register tx                            setMetadata
            │                                      │
            ▼                                      ▼
   ┌────────────────────┐               ┌────────────────────┐
   │   WRITE upstream   │               │   READ upstream    │
   │   (eRPC primary)   │               │  (eRPC fallback,   │
   │                    │               │   public RPC, OR a │
   │   Register lands   │               │   stale Anvil fork │
   │   in block N       │               │   pinned by a prior│
   │                    │               │   flow-12/13 run)  │
   └────────────────────┘               └────────────────────┘
            │                                      │
            └─── WaitMined ✓                       └── eth_estimateGas
                                                       at "latest", which
                                                       on the reader side is
                                                       still block N-1 OR a
[...]

PR #386, opened 2026-04-27
Integrate PR #377 (OBOL Permit2) + PR #381 (Hermes runtime), harden flows, add flow-13 dual-stack OBOL

bussyjd:
## Summary

This PR integrates **#377 (OBOL Permit2)** and **#381 (Hermes default agent runtime)** into a single testable branch and lands the hardening needed to keep both flows green together against real infrastructure (live Base Sepolia, real Cloudflare tunnel, two independent obol stacks on the same host).

Two end-to-end test flows now pass cleanly:

| Flow | Asset | Chain | Result |
|------|-------|-------|--------|
| `flow-11-dual-stack.sh` | USDC | live Base Sepolia (84532) | **48/45 PASS, 0 FAIL** |
| `flow-13-dual-stack-obol.sh` (new) | OBOL ERC20Permit | Anvil fork of 84532 | **56/53 PASS, 0 FAIL** |

PASS-count > step-count because some steps assert multiple invariants (e.g. step 21 archives 2 receipts + 1 Agent ID).

---

### flow-11 — live Base Sepolia, USDC, ERC-8004

```
                    ┌─────────────────────────────────────────────┐
                    │   Base Sepolia (live, public)               │
                    │     ERC-8004 IdentityRegistry: 0x8004A818…  │
                    │     USDC:                       0x036CbD53… │
                    │     Facilitator: x402.gcp.obol.tech         │
                    └────────────▲─────────────────────▲──────────┘
                                 │                     │
                  ┌──── register/metadata tx ────┐    │
                  │                              │    │
                  │              ┌── Cloudflare ─┘    │
                  │              │  (real public      │
                  │              │  *.trycloudflare.com)
                  │              │                    │ USDC
                  │              ▼                    │ Transfer
                  │     ┌─────────────────────┐      │ (settlement)
                  │     │  cloudflared (k3d)  │      │
                  │     └────────┬────────────┘      │
                  │              │                    │
       ┌──────────┴──────────┐   │   ┌────────────────┴────────────┐
       │  Alice k3d cluster  │   │   │  Bob k3d cluster            │
       │  ─────────────────  │   │   │  ───────────────            │
       │  Traefik gateway    │   │   │  Hermes agent (api server)  │
       │  ServiceOffer CR    │◄──┘   │  buy.py (x402)              │
       │  LiteLLM + Ollama   │       │  x402-buyer sidecar         │
       │  remote-signer      │       │  remote-signer (signs auths)│
       └─────────────────────┘       └─────────────────────────────┘
[...]

PR #391, opened 2026-04-29
refactor(network): drop chain-id reachability probe from `obol network status`

OisinKyne:
The probe (introduced via #388) was sold as detecting "stale custom RPC          
  pins from a previous test run", but the symptom it was built to chase                                                        
  — eRPC returning ERC721NonexistentToken on a freshly-minted agent ID
  — was actually read-side staleness in eRPC, not a stale Anvil fork.                                                          
  PR #387's `Client.WaitForAgent` is the correct fix and is already                
  wired into the CLI register paths.                                                                                           
                                                                  
  eth_chainId is also a poor liveness probe: it can't catch lagging                                                            
  upstreams, and an Anvil fork of base-sepolia returns the same chain              
  id (84532) as the real chain — exactly the case it was built to                                                              
  detect.                                                                                                                      
                                                                                                                               
  eRPC already tracks per-upstream health metrics; if we want a future                                                         
  `obol network status` reachability column it should parse those, not                                                         
  redo the work badly in-process.                                                                                              
                                                                                                                               
  Changes:                                                                                                                     
  - internal/network/probe.go, probe_test.go: deleted.                                                                         
  - cmd/obol/network.go: drop --no-probe / --probe-timeout flags,                                                              
    renderUpstreamProbes, uiPrinter, and the `time` import.                                                                    
                                                                                   
  flow-02 and flow-05 still pass: they grep `eRPC|Pod|Upstream` headers                                                        
  which remain in the output.                                                                                                  
                                                                                                                               
  Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


bussyjd:

PR #392, opened 2026-04-29
refactor(stack): drop Helm-ownership migration hacks

OisinKyne:
The 9ef8cda commit added two pieces of compatibility ballast that
  ran on every `obol stack up`:                                                                                                
                                                                                   
  1. migrateBaseHelmOwnership — labelled and annotated a hardcoded list
     of pre-existing namespaces/RBAC objects to hand them to Helm.                                                             
     The trigger was a one-off transition where prior obol versions
     created those objects outside Helm; brittle (any new resource                                                             
     needs another entry) and only exists to paper over a known dev-                                                           
     machine state.                                               
                                                                                                                               
  2. needsLiteLLMConfigHelmMigration / the conditional delete-and-                 
     restore branch in preserveLiteLLMConfigForHelm — fired when the                                                           
     litellm-config ConfigMap had a non-`helm` field manager                                                                   
     (typically `kubectl-patch` from old `obol model setup` calls).                                                            
     Current model.go patches use --field-manager=helm explicitly,                                                             
     so the trigger is unreachable in the current code path.                                                                   
                                                                                                                               
  Both fall under "code for a one-off issue that already passed".                                                              
                                                                                                                               
  What stays: preserveLiteLLMConfigForHelm (simplified to always                                                               
  snapshot), restoreLiteLLMConfig, mergeLiteLLMConfig, and                                                                     
  ApplyServerSideForceConflicts. The base helmfile template only                                                               
  includes the `paid/*` catch-all route; without preserve+restore,                                                             
  every `obol stack up` would wipe user-added cloud providers and                                                              
  custom endpoints. That's a real-resilience exception, not a hack.                                                            
                                                                                                                               
  Tests: TestNeedsLiteLLMConfigHelmMigration removed.                                                                          
  TestMergeLiteLLMConfigPreservesChartDefaultsAndPreviousModels stays.
                                                                                                                               
  Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/internal/stack/stack.go
+++ b/internal/stack/stack.go
@@ -350,10 +350,6 @@ func syncDefaults(cfg *config.Config, u *ui.UI, kubeconfigPath string, dataDir s
 	defaultsHelmfilePath := filepath.Join(cfg.ConfigDir, "defaults")
 	helmfilePath := filepath.Join(defaultsHelmfilePath, "helmfile.yaml")
 
-	if err := migrateBaseHelmOwnership(cfg, kubeconfigPath); err != nil {
-		u.Warnf("Failed to migrate existing base resources into Helm ownership: %v", err)
-	}
-
 	previousLiteLLMConfig, err := preserveLiteLLMConfigForHelm(cfg, kubeconfigPath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bussyjd:

PR #400, opened 2026-04-30
fix(stack): reuse cached dev images on stack up

bussyjd:
## Summary
- keep `internal/stack` default dev semantics deterministic: source-backed local images still rebuild on normal `obol stack up`
- add an explicit opt-in cache-reuse path via `OBOL_REUSE_LOCAL_DEV_IMAGES=true`
- have `flows/flow-11-dual-stack.sh` opt into that behavior because it is a smoke/validation harness rather than a source-editing loop
- add regression tests for both behaviors: default rebuild vs opt-in cache reuse

## Why
The original stall I hit on latest `main` happened during Alice `stack up` in `flow-11`: `buildAndImportLocalImages()` tried to rebuild `ghcr.io/obolnetwork/x402-verifier:latest` even though that tag was already present locally, and the rebuild blocked while resolving `golang:1.25-alpine`.

I agree that making cache reuse the default in `internal/stack/stack.go` felt too smoke-test-driven and could hide fresh local source changes during normal development. This revision keeps the product-path default unchanged and moves the smoke-specific policy to the smoke harness.

## Design
- `internal/stack/stack.go`
  - add `OBOL_REUSE_LOCAL_DEV_IMAGES=true` as an explicit opt-in
  - when set, reuse already-cached local/preload images if present
  - otherwise keep current behavior: rebuild local repo images / pull preload refs
- `flows/flow-11-dual-stack.sh`
  - exports `OBOL_REUSE_LOCAL_DEV_IMAGES=true` by default for that flow only

## Test Plan
- `bash -n flows/flow-11-dual-stack.sh`
- `go test ./internal/stack -run 'TestBuildAndImportLocalImages_(DefaultBuildsEvenWhenImageExists|ReusesExistingImageWhenOptedIn)' -count=1`
- `go test ./internal/hermes -run 'TestImportPrivateKeyWalletCmd_(ApplyClusterTrueRollsPod|ApplyClusterFalseSkipsCluster|SyncFailureSkipsRestart)' -count=1`
- `go test ./cmd/obol ./internal/stack ./internal/hermes -count=1`
- manual smoke: reran the `flow-11` setup path from this branch and confirmed the earlier `docker build ... Dockerfile.x402-verifier` stall no longer appeared; with the flow opting into cache reuse, Alice `stack up` advanced to a cold `docker pull nousresearch/hermes-agent:v2026.4.23` instead of hanging on the verifier rebuild

### DIFF
--- a/flows/flow-11-dual-stack.sh
+++ b/flows/flow-11-dual-stack.sh
@@ -35,6 +35,11 @@
 #   FLOW11_BOB_HTTPS_PORT  FLOW11_BOB_HTTPS_ALT_PORT
 source "$(dirname "$0")/lib.sh"
 
+# This flow is a smoke/validation harness, not a source-editing loop. Reuse
+# already-cached local dev images when available so repeated runs don't block on
+# unnecessary rebuilds or cold pulls during `obol stack up`.
+export OBOL_REUSE_LOCAL_DEV_IMAGES="${OBOL_REUSE_LOCAL_DEV_IMAGES:-true}"
+
 # ═════════════════════════════════════════════════════════════════
 # PREFLIGHT
 # ═════════════════════════════════════════════════════════════════
--- a/internal/stack/stack.go
+++ b/internal/stack/stack.go
@@ -623,8 +623,8 @@ func devPreloadImages() []string {
 	return images
 }
 
-func forceRebuildLocalImages() bool {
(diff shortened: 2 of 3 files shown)


bussyjd:
Superseded by #401, which includes this branch's cached-image reuse commits plus the live OBOL smoke, cloudflared, Renovate, and skill updates in one PR.

bussyjd:
Closing in favor of #401 so the review happens in a single combined PR.

PR #411, opened 2026-05-04
Hermes recreation issues

OisinKyne:
## Summary

Not sure if this works. Had trouble with hermes recreates after stack down and up. Opus wants to strip most of the perms messing and custom image building. I'm 60% confident that this is the right and complete fix.


bussyjd:
Superseded by #413 for the remaining actionable parts of this PR. The replacement keeps the Hermes image-based runtime simplification, bumps to `nousresearch/hermes-agent:v2026.4.30`, adds an init preflight for the image-provided binary/extras, fixes stale Hermes wallet metadata on restore, and leaves the LiteLLM ConfigMap delete/recreate behavior out as a separate decision unless `main` still reproduces the issue.

OisinKyne:
Closed by #413

PR #422, opened 2026-05-05
feat(model): auto-discover local OpenAI-compatible inference servers

bussyjd:
## Summary

Extends `obol stack up` auto-config beyond Ollama to register vLLM, sglang, llama.cpp server, LM Studio, mlx-lm, and any other OpenAI-compatible server detected on a well-known local port — operators no longer have to run `obol model setup custom --endpoint ...` per server.

- **`internal/model/discover.go`** — thin adapter over `inference.ScanLocalEndpointsContext` that filters Ollama (handled by the existing `ListOllamaModels` path) and LiteLLM (would be self-referential), then converts each `EndpointInfo` to `ModelEntry` values via the existing `buildCustomEndpointEntry` + `localhostToClusterEndpoint` helpers.
- **`internal/inference/detect.go`** — adds LM Studio (`:1234`) to the default probe list, exports `ServerType*` and `LiteLLMPort` constants so callers match by symbol, and supports `OBOL_LOCAL_DISCOVERY_PORTS=port[:label],...` to add custom ports.
- **`internal/stack/stack.go`** — `autoConfigureLLM` calls `autoDetectLocalProviders` and batches **all** discovered providers into a **single** ConfigMap PATCH (atomic, one round-trip).
- **`cmd/obol/model.go`** — new `obol model discover` subcommand (read-only) and extended `obol model status` (human + JSON) that surface what auto-config picked up.
- **`internal/model/model.go`** — `PatchLiteLLMProvider` delegates its tail to a new exported `PatchLiteLLMEntries(cfg, u, entries)` helper.

`OBOL_DISABLE_LOCAL_DISCOVERY=1` opt-out for cases where a port collides with an unrelated service.

### Interaction with `obol model prefer`

This PR appends discovered servers **after** the Ollama branch in `autoConfigureLLM`, so Ollama keeps default primary-model status under the new "model_list order is source of truth" `Rank()` contract. Operators who want vLLM (or any discovered server) as primary can use `obol model prefer <model>` after stack up.

## Test plan

- [x] `go build ./...` clean
- [ ] `go vet ./...` clean *(pre-existing macOS baseline failure in `internal/enclave/enclave_darwin.go`; `go vet ./internal/model ./internal/hermes ./internal/stack` passes)*
- [x] `go test ./...` passes (full suite)
- [x] New unit tests:
  - `internal/model/discover_test.go` — `shouldRegisterEndpoint` filters (LiteLLM port, Ollama, no-models), `buildDiscoveredProvider` translates `127.0.0.1` → `host.k3d.internal`, env-var disable
  - `internal/inference/detect_test.go` — `extraPortsFromEnv` parsing edge cases, `resolvedProbePorts` dedup keeps default labels
- [x] Manual: `obol model discover` on a host with vLLM/sglang/LM Studio running prints the expected providers *(validated on spark1 DGX Spark host: `llama-server@8000` with `qwen36-35b-heretic`, `qwen36-fast`, `qwen36-deep`)*
- [x] Manual: `obol stack up` on the same host shows discovered servers registered in `obol model list` without a manual `obol model setup custom` *(validated on spark1: gateway `openai` provider contained the three discovered qwen36 models after stack up)*
- [x] Manual: `OBOL_DISABLE_LOCAL_DISCOVERY=1 obol stack up` skips the scan entirely *(validated on spark2 with a live mock `:9000`; resulting `obol model list` left `openai` disabled and omitted `spark9000-probe-a/b`)*
- [x] Manual: `OBOL_LOCAL_DISCOVERY_PORTS=9000:vllm` adds a custom port to the scan *(validated on spark2 mock `:9000`; `obol model discover` and post-`stack up` `obol model list` both included `spark9000-probe-a/b`)*

## Out of scope

- No vllm-studio integration. The repo (https://github.com/0xsero/vllm-studio) inspired the multi-provider discovery direction; this PR implements a Go-native equivalent that fits the thin-CLI architecture.
- No HuggingFace cache scanning — probing live `/v1/models` endpoints is the right unit of discovery.
- No SIGINT propagation through the CLI (pre-existing; `cmd/obol/main.go` runs with `context.Background()`).

https://claude.ai/code/session_01K6hF5YYtT46VbNJbRyGy4Q

---
_Generated by [Claude Code](https://claude.ai/code/session_01K6hF5YYtT46VbNJbRyGy4Q)_


bussyjd:
Reviewer feedback on rc8 (post-upgrade) reports a related model-selection regression that this PR's discovery path will need to handle:

> After the upgrade, regenerated `values-hermes.yaml` set `default: deepseek-v4-pro:cloud` — a model that requires an API key the user almost certainly hasn't configured. Hermes will fail to respond to chat until the user manually edits to a local model.
>
> History on this machine across recent RCs:
> - rc1 → `llama3.2:1b` (good)
> - rc3 → `qwen3.5:4b` (slow on CPU)
> - rc7 → `llama3.2:1b` (good)
> - rc8 → `deepseek-v4-pro:cloud` (unusable without API key)

Root cause is in `internal/model/rank.go` — `Rank()` picks `models[0]` from the LiteLLM `model_list` verbatim ("model_list order is the source of truth"), so whatever ends up at index 0 wins. On this machine an Ollama-pulled `*:cloud` entry (cloud-suffixed model that Ollama lists but cannot serve without an API key) landed first and was promoted to primary.

This PR appends discovered providers **after** the Ollama branch, which is fine — but the Ollama branch itself can still hand `Rank()` a `:cloud`-suffixed entry as `models[0]`. Worth considering as part of this PR (or a follow-up):

- Filter `*:cloud` (and any other "requires-API-key" Ollama suffix) out of the auto-config model list when no key is present, **or**
- Demote them to fallback in `Rank()` the same way embedding-only models are demoted, **or**
- At minimum, surface a warning during `obol stack up` when the auto-picked primary requires an unconfigured credential.

Suggested fix the reviewer offered:

> prefer the smallest tool-capable local Ollama model on machines without a configured cloud provider, or prompt the user during `obol stack up` if multiple candidates exist.

OisinKyne (approved):

PR #428, opened 2026-05-05
feat(buyer): if a user requests html, give them html

OisinKyne:
## Summary

Creates HTML responses for x402 service endpoints. Returns json if content-type is Accept: application/json, or not set, if text/html, it returns this site. This allows us to give link previews and guides on the service page itself if it works out.


OisinKyne:
@bussyjd you can check i haven't broken anything too hard in the morning before we pull the publish trigger. Kody may be a pass today

PR #432, opened 2026-05-06
fix(defaults): resolve host gateway via Docker for Colima/Rancher Desktop

apham0001:
## Summary

`OllamaHostIPForBackend` hardcoded the Docker Desktop magic gateway IP (`192.168.65.254`) on `darwin+k3d`. On **Colima** and **Rancher Desktop** the host gateway lives on a different bridge (e.g. `192.168.5.2` for Colima's default profile), so the rendered Ollama `Endpoints` pointed at an IP no pod could reach. The Service silently timed out for every LiteLLM call, surfaced upstream in OpenClaw as the cryptic:

> `provider rejected the request schema or tool payload`

This PR resolves the host gateway by running a transient `alpine:3` container that prints `host.docker.internal` from inside Docker — works the same on Docker Desktop, Colima, and Rancher Desktop because all three expose `host.docker.internal` inside containers (mapping it to whatever bridge IP their VM uses).

Resolution order on `darwin+k3d`:
1. **Docker-based lookup** (new): `docker run --rm alpine:3 getent hosts host.docker.internal` → falls back to `--add-host=host.docker.internal:host-gateway` for stricter setups.
2. Docker Desktop magic IP (`192.168.65.254`) as a last resort, so installs without a working `docker` CLI still behave as before.
3. Linux unchanged (`docker0`/`br-*` bridge fallback).

Also adds a **"Minimum local model size"** note to the README. The stack relies heavily on tool-calling, and 1B–4B / `*-coder` Ollama variants either return raw JSON in `content` or hallucinate tool failures. Recommends `llama3.1:8b` / `qwen3:8b` / `qwen2.5:7b` (instruct) for the agent role.

## Why this happened

`net.LookupHost("host.docker.internal")` runs on the **host machine**, where the name is not resolvable (it only exists inside Docker containers). The function therefore always fell through to `DockerDesktopGatewayIP()`. On Docker Desktop that IP happens to be correct; on Colima/Rancher it's wrong, and the cluster gets a black-hole Endpoint.

## Test plan

- [x] `go test ./...` (unit suite, full repo) — all green
- [x] `go vet ./internal/defaults/...` — clean
- [x] `gofmt -l` — clean
- [x] Smoke test on **Colima** (default profile, 4 CPU / 8 GiB, k3d backend, macOS Apple Silicon):
  - `ResolveHostGatewayViaDocker()` returns `192.168.5.2` in ~400ms
  - `OllamaHostIPForBackend("k3d")` returns `192.168.5.2`
  - Rendered `Endpoints` becomes reachable from in-cluster pods
  - OpenClaw chat completes a full LiteLLM → Ollama → tool-call round-trip with `llama3.1:8b`
- [ ] **Verification on Docker Desktop pending** — needs a reviewer with Docker Desktop to confirm the new code path returns `192.168.65.254` (the magic IP fallback is still in place, so behaviour should be identical even if Docker resolution fails)
- [ ] **Verification on Linux/Rancher Desktop pending**

## Notes

- New container spawn at `obol stack init` adds ~400 ms of latency. Acceptable; only happens at init/refresh.
- `alpine:3` is auto-pulled if missing (~7 MB).
- `DockerDesktopGatewayIP()` is preserved as a public symbol with an updated doc comment, since `internal/stack/stack.go` still re-exports it.
- README addition is concise and kept in the "Model Providers" section to be visible during initial setup, where the choice matters most.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,17 @@ Find charts at [Artifact Hub](https://artifacthub.io).
 
 ## Model Providers
 
-The stack runs [LiteLLM](https://github.com/BerriAI/litellm) as an in-cluster OpenAI-compatible gateway that proxies all LLM traffic. By default, Ollama on the host machine is used. To switch to a cloud provider:
+The stack runs [LiteLLM](https://github.com/BerriAI/litellm) as an in-cluster OpenAI-compatible gateway that proxies all LLM traffic. By default, Ollama on the host machine is used.
+
+**Minimum local model size**: agents in this stack rely heavily on tool calling (skills are exposed as OpenAI-style tools, the agent's identity bootstrap reads files, etc.). Models below ~7B parameters tend to either ignore the structured tool-calling channel and return raw JSON in the assistant message, or hallucinate tool failures without actually invoking the tool. Recommended local minimums for reliable agent behaviour:
+
+- `llama3.1:8b` — reference Ollama tool-calling implementation
+- `qwen3:8b` — strong tool support, modern
+- `qwen2.5:7b` (instruct) — works, slightly less disciplined under load
+
+Avoid the 1B–4B and `*-coder` variants for the agent role — they pass simple chats but break under multi-tool workflows. They remain fine for embeddings or single-turn completions exposed via `obol sell inference`.
+
+To switch to a cloud provider:
 
 ```bash
 # Interactive — prompts for provider and API key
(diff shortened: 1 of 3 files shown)


bussyjd (approved):

OisinKyne:
lgtm, won't merge pre 0.9.0 as idk how many non docker desktop people there are on mac. can go in after. we can also tip docker desktop users to increase their resources if they have the defaults which are too tight

PR #437, opened 2026-05-06
Hanan/fix bugs

HananINouman:
Title: fix: registration blocking, stale images, and skill catalog
  self-healing

  Body:

  Three bugs found and fixed during v0.9.0-rc11 fresh-cluster testing.

  ---
  Bug 1 — Unhealthy service blocks ERC-8004 registration ownership

  internal/serviceoffercontroller/controller.go — registrationOffers()

  registrationOffers() never checked upstream health when selecting registration
   owner candidates. A service with a failing upstream (e.g. Ollama down) could
  become the oldest candidate and block every newly created service from ever
  reaching Registered.

  Fix: Skip offers where UpstreamHealthy is not True.

  ---
  Bug 2 — Stale images silently run old code (OBOL/USDC mismatch, missing
  features)

  internal/embed/infrastructure/base/templates/x402.yaml

  Both containers had imagePullPolicy: IfNotPresent. k3d's containerd cache is
  separate from the Docker daemon — docker pull never updates it. Pods reused
  old cached images indefinitely after restart, causing x402-verifier to
  advertise USDC for OBOL offers, and serviceoffer-controller to run a build
  that predated services.json.

  Fix: imagePullPolicy: Always on both containers.

  ---
  Bug 3 — Skill catalog stuck in ContainerCreating; /api/services.json empty

  internal/serviceoffercontroller/controller.go

  Two compounding issues:
  1. SSA field-manager trap: old cluster state had a stale api-content volume on
   the Deployment owned by kubectl-patch. SSA with Force: true cannot remove
  volumes owned by a foreign field manager, so every reconcile left the stale
  volume. Pod failed ContainerCreating looking for a ConfigMap that no longer
  exists.
  2. No self-healing: skill catalog only reconciled on ServiceOffer events —
  external deletions were never repaired.

  Fix:
  - deleteSkillCatalogDeploymentIfStale() — detects unexpected volumes and
  deletes the Deployment before re-applying, since SSA can't do an in-place
  removal.
  - 30-second periodic ticker in Run() — skill catalog self-heals within 30s
  without needing a ServiceOffer event.

  Note: no nginx needed. /api/services.json is served by the existing busybox
  httpd via a ConfigMap items projection.

  ---
  Test plan
  - Fresh cluster: /api/services.json returns services array after obol sell
  - Upgrade from old cluster with stale api-content volume: Deployment deleted
  and recreated cleanly
  - Service with Ollama down: not selected as registration owner; healthy
  service claims it instead
  - New :latest published: pods pull updated image on restart without manual k3d
   image import

### DIFF
--- a/internal/embed/embed_image_pin_test.go
+++ b/internal/embed/embed_image_pin_test.go
@@ -33,9 +33,7 @@ func TestEmbeddedImages_NoNewLatestTags(t *testing.T) {
 	// Known unpinned images as of PR #343 follow-up. Each entry MUST have a
 	// TODO in the template body explaining the pin-by-digest policy.
 	allowed := map[string]string{
-		"base/templates/llm.yaml:ghcr.io/obolnetwork/x402-buyer:latest":               "x402-buyer: pin by digest once CI publishes a stable tag",
-		"base/templates/x402.yaml:ghcr.io/obolnetwork/x402-verifier:latest":           "x402-verifier: pin by digest once CI publishes a stable tag",
-		"base/templates/x402.yaml:ghcr.io/obolnetwork/serviceoffer-controller:latest": "serviceoffer-controller: pin by digest once CI publishes a stable tag",
+		"base/templates/llm.yaml:ghcr.io/obolnetwork/x402-buyer:latest": "x402-buyer: pin by digest once CI publishes a stable tag",
 	}
 
 	files := []string{
--- a/internal/embed/infrastructure/base/templates/x402.yaml
+++ b/internal/embed/infrastructure/base/templates/x402.yaml
@@ -186,7 +186,7 @@ spec:
           # masks protocol regressions (e.g. facilitator payload format, verifyOnly
           # handling). See internal/embed/embed_image_pin_test.go.
           image: ghcr.io/obolnetwork/x402-verifier:latest
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
           ports:
             - name: http
               containerPort: 8080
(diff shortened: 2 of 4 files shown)


OisinKyne:
> Fix: imagePullPolicy: Always on both containers.
too aggressive for non developers, can't always be re-downloading stuff. better to switch to commit shas and then as they change, new code is pulled once not every single time.

OisinKyne:
> Fix:
> 
>     deleteSkillCatalogDeploymentIfStale() — detects unexpected volumes and
>     deletes the Deployment before re-applying, since SSA can't do an in-place
>     removal.
>     30-second periodic ticker in Run() — skill catalog self-heals within 30s
>     without needing a ServiceOffer event.
> 
> Note: no nginx needed. /api/services.json is served by the existing busybox
> httpd via a ConfigMap items projection.
bit unsure of this too. imo we're nearly off the bad states (jd and i had similar issues). I think maybe we should just jump to helm 4 and use the new features jd pulled out recently. running deletes every 30 seconds of the day to fix a bug doesn't sound right to me.

OisinKyne:
> internal/serviceoffercontroller/controller.go — registrationOffers()
> 
> registrationOffers() never checked upstream health when selecting registration
> owner candidates. A service with a failing upstream (e.g. Ollama down) could
> become the oldest candidate and block every newly created service from ever
> reaching Registered.
> 
> Fix: Skip offers where UpstreamHealthy is not True.

this i don't understand but maybe support. maybe this is coupling registrationOffers too tightly to inference only, which is what it was originally programmed for? but maybe this means don't show unhealthy services in the ui. Which maybe i support but maybe we just expose with some red status indicators or something

HananINouman:
> > Fix: imagePullPolicy: Always on both containers.
> > too aggressive for non developers, can't always be re-downloading stuff. better to switch to commit shas and then as they change, new code is pulled once not every single time.

maybe until the TODO (image-pin) is done?

OisinKyne:
> maybe until the TODO (image-pin) is done?
this only impacts pre-existing users, so i think thats basically just the core team, and i'd rather inconvenience them versus add nearly a minute to every startup.

OisinKyne (approved):

PR #446, opened 2026-05-07
fix: hermes PVC ownership on Linux k3d + bootstrap ingress URL

nickh-obol:
## Summary

Two Linux-specific bugs in fresh `obol stack up` flows that don't surface on macOS.

### 1. Hermes pod CrashLoopBackOff on fresh PVC (`internal/hermes/hermes.go`)

The Stack's local-path-provisioner setup script chowns each new PV to `1000:1000`, and `KubeletInUserNamespace=true` (set in `internal/embed/k3d-config.yaml`) silently skips the `fsGroup` recursive-chown that would otherwise correct it. Hermes pods run as UID 10000 and crashloop trying to `mkdir /data/.hermes/home` with `Permission denied`.

macOS Docker Desktop hides this — its bind-mount filesystem driver fakes file ownership to whoever's asking, so the mismatch never materializes. OpenClaw doesn't hit it either: it pins `fsGroup: 1000`, accidentally matching the host UID.

**Fix:** prepend an `init-hermes-perms` container that runs as root and chowns `/data` to 10000:10000. Idempotent — also self-heals existing broken PVCs on upgrade.

### 2. Bootstrap probe + browser URL hardcoded to `:8080` (`cmd/obol/bootstrap.go`)

When ports 80/8080 are in use at start time, k3d remaps the loadbalancer to a random high port and the rest of `obol stack up` correctly surfaces it via `stack.LocalIngressURL(cfg)`. Three other places in `cmd/obol/bootstrap.go` were hardcoded to `:8080`: the readiness probe (hung until timeout), the browser-open URL, and the "view stack interface at" hint.

**Fix:** reuse `stack.LocalIngressURL(cfg)` in all three places.

## Test plan

- [x] `go test ./internal/hermes/... ./cmd/obol/...` passes
- [x] `go build ./...` clean
- [x] Verified end-to-end on Linux: fresh `obol agent init` lands the agent in `Running` without manual chown
- [x] Verified end-to-end on Linux: `obol bootstrap` with port 8080 occupied completes without hang; all four URL surfaces (warning / visit / browser-open / next-steps) agree on the alternate port
- [x] Smoke-test on macOS to confirm no regression (existing fsGroup path still works; new init container is a no-op on already-correct ownership)


bussyjd (approved):
Validated against latest origin/main in an isolated worktree.

Checks run locally:
- git merge --no-commit --no-ff refs/remotes/pr/446/head onto origin/main
- git diff --check --cached
- go test ./internal/hermes ./internal/stack ./cmd/obol
- go build ./cmd/obol
- go test ./...

All passed. The Hermes PVC ownership fix and bootstrap LocalIngressURL wiring look legitimate.

PR #497, opened 2026-05-17
fix(sell): allow concrete paid/<model> as default agent chat model

bussyjd:
## Summary

- `resolveDefaultAgentModel` was skipping ALL `paid/`-prefixed entries via `strings.HasPrefix`; only the literal `paid/*` wildcard (a LiteLLM routing namespace entry) should be filtered out
- Extracts a pure `pickAgentDefault([]string)` helper and adds 5 table-driven unit tests covering the key cases

## Why

After `obol buy inference v1337-aeon --model aeon && obol model prefer paid/aeon`, running `obol sell demo quant` would pin the agent to the **second-rank** model because all `paid/*`-prefixed entries were being skipped. The wildcard `paid/*` and a concrete `paid/aeon` are different LiteLLM concepts: the wildcard is a routing namespace (not callable), while concrete purchased entries route through the x402-buyer sidecar and are fully valid chat models. This fix makes "agent uses my preferred model" work without a `--chat-model` flag.

## Test plan

- [x] `TestPickAgentDefault` — 5 cases: concrete-paid-at-head wins, wildcard-skipped-fallthrough, wildcard-only error, concrete-paid-only, empty-list error
- [ ] Manual: `obol buy inference <name> --model <id>` → `obol model prefer paid/<id>` → `obol sell demo quant` → `kubectl get agent quant -n agent-quant -o jsonpath='{.spec.model}'` should return `paid/<id>`

### DIFF
--- a/cmd/obol/sell_agent.go
+++ b/cmd/obol/sell_agent.go
@@ -427,9 +427,14 @@ func runAgentBackedDemo(
 
 // resolveDefaultAgentModel picks a model to pin onto a fresh Agent CR.
 // Walks the cluster's LiteLLM model_list (the same source `obol model
-// list` reads), drops the meta `paid/*` route, and returns the top
-// entry. The list is already in the operator's preferred order via
-// `obol model prefer`, so "first non-paid" is a meaningful default.
+// list` reads), drops the wildcard `paid/*` meta route, and returns the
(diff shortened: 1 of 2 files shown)


OisinKyne (approved):
What is the point of `paid/*` so? do we need to keep it at all?

PR #548, opened 2026-05-25
fix(monetize): drop available, use drainEndsAt as sole drain signal

bussyjd:
## Summary

drain becomes purely additive — active offers serialize identically to pre-drain main. The only new wire field is `drainEndsAt`, set on draining offers only. Consumers detect drain with `if (entry.drainEndsAt) { /* draining */ }`. No schema-breaking change for any consumer that was reading the catalog before drain landed.

This re-ships an amendment that was originally pushed as commit dd89750 on `feat/drain-replaces-pause` for PR #535. The amendment didn't survive the bundle PR #536's merge into main, so the controller is shipping the un-amended `Available bool` shape today.

## Diff summary

- `internal/schemas/service_catalog.go` — remove `Available bool` from `ServiceCatalogEntry`; keep `DrainEndsAt string `omitempty\`` as the sole drain marker.
- `internal/schemas/service-catalog.schema.json` — drop `available` from `required` and from `properties`; tighten the `drainEndsAt` description to "present iff draining".
- `internal/serviceoffercontroller/render.go`
  - `buildServiceCatalogJSON`: stop setting `Available`; only set `DrainEndsAt` when `offer.IsDraining()`.
  - `buildSkillCatalogMarkdown`: rename the `Available` table column to `Status` (active rows show `—`; draining rows show ``draining · ends `<RFC3339>` ``); drop the per-service `- **Available**:` bullet entirely (draining services keep only `- **Drain ends at**:`).
  - `serviceDefWithDrain`: stop setting the (already-additive) `Available *bool` on `erc8004.ServiceDef` during drain — signal drain via `DrainEndsAt` only.
- `internal/serviceoffercontroller/render_test.go`
  - `TestBuildServiceCatalogJSON_ExcludesNonReady`: replace `services[0].Available == true` assertion with a raw-JSON map walk asserting `available` and `drainEndsAt` keys are absent on active entries.
  - `TestBuildServiceCatalogJSON_DrainLifecycle`: rewrite the lifecycle assertions to walk `[]map[string]any`. Active entries: neither `available` nor `drainEndsAt`. Mid-drain: only `drainEndsAt`. Expired: filtered.
  - `TestBuildRegistrationServices_IncludesDrainMetadata` and `TestBuildIdentityRegistrationServices_IncludesDrainMetadata`: replace `svc.Available == &false` checks with `svc.Available == nil` (the helper no longer sets it; presence-of-`DrainEndsAt` is the sole drain marker).
  - **NEW** `TestBuildSkillCatalogMarkdown_DrainAdditiveDetail`: asserts no `- **Available**:` bullet appears for any offer, draining offers keep their `- **Drain ends at**:` bullet, and the table header uses `Status` not `Available`.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/schemas/... ./internal/serviceoffercontroller/... ./internal/monetizeapi/...` green
- [x] `go test ./...` green (full suite, including `internal/stack`)
- [x] `TestBuildServiceCatalogJSON_ExcludesNonReady` confirms active-offer JSON has no `available` key
- [x] `TestBuildServiceCatalogJSON_DrainLifecycle` confirms draining offer still emits `drainEndsAt` (and only `drainEndsAt`)
- [x] `TestBuildSkillCatalogMarkdown_DrainAdditiveDetail` confirms markdown surface is additive: no `- **Available**:` bullet, `Status` column header, `- **Drain ends at**:` retained on draining offers.

### DIFF
--- a/internal/schemas/service-catalog.schema.json
+++ b/internal/schemas/service-catalog.schema.json
@@ -78,8 +78,7 @@
         "payTo",
         "network",
         "description",
-        "isDemo",
-        "available"
+        "isDemo"
       ],
       "properties": {
         "name": {
@@ -154,14 +153,10 @@
         "registrationPending": {
           "type": "boolean"
         },
-        "available": {
-          "type": "boolean",
-          "description": "False during a drain window. Catalog consumers should treat unset as true for backwards compatibility."
-        },
         "drainEndsAt": {
           "type": "string",
           "format": "date-time",
-          "description": "RFC3339 timestamp at which the offer's HTTPRoute will be torn down. Set only when available=false."
+          "description": "RFC3339 timestamp at which the offer's HTTPRoute will be torn down. Set only when the offer is draining. Catalog consumers should detect drain via the presence of this field."
         }
       }
     }
--- a/internal/schemas/service_catalog.go
+++ b/internal/schemas/service_catalog.go
@@ -40,19 +40,12 @@ type ServiceCatalogEntry struct {
 	// ERC-8004 discovery via the chain still resolves to the prior state.
 	RegistrationPending bool `json:"registrationPending,omitempty"`
 
-	// Available is false when the offer is in its drain window. Buyers
-	// can still complete in-flight payments until DrainEndsAt, but
-	// discovery surfaces should advertise the wind-down so external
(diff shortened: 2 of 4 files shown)


OisinKyne:
Superseeded. by #551

PR #587, opened 2026-06-03
chore(dev): tag dev images :dev-<sha> instead of :latest (no cross-worktree poisoning)

bussyjd:
## Summary

What changed: Under `OBOL_DEVELOPMENT`, locally-built images are now tagged
`<image>:dev-<short-git-sha>` instead of the shared, mutable `<image>:latest`. The
manifest rewrite (`internal/defaults`) and the image build (`internal/stack`) agree on
the tag via a persisted `$CONFIG_DIR/.dev-image-tag`.

Why it matters: `:latest` is shared across every obol-stack worktree on one Docker
daemon. One worktree's build silently overwrites `:latest`, and another worktree's
`obol stack up` then deploys the wrong binary. This bit rc9 QA: a sibling branch's
`serviceoffer-controller` (which reads a per-purchase Secret this branch's RBAC doesn't
grant) poisoned an unrelated stack and presented as a buy hang, not an obvious image
mismatch. A per-commit `dev-<sha>` tag isolates builds by source so this can't recur.
Bonus: committing changes the SHA and auto-triggers a fresh build.

Risk level: low — dev-only path (production digest pins untouched); `:latest` fallback
preserves behaviour for non-git/tarball builds; the build loop is otherwise unchanged
(only the tag string differs).

Commit under test: `7b9c479d`

Base branch: `main`

## Scope

- [x] Code
- [ ] Charts / manifests
- [x] Flows / QA scripts 
- [ ] Docs / skills
- [x] Images / dependencies 
- [ ] Other:

## Validation

Unit tests:

```text
go test ./...   → ok (33 packages), gofmt + go vet clean   (commit 7b9c479d)
New/updated:
  internal/defaults  TestDevImageTag_Format, TestReadDevImageTag_FallbackWhenAbsent
  internal/defaults  TestCopyInfrastructure_DevModeRewritesDigestPins (now asserts dev-<sha>
                     + that the persisted tag matches the stamped manifest tag)
```

Manual / runtime:

```text
OBOL_DEVELOPMENT=true obol stack init  →
  .dev-image-tag = dev-d9527983fc74
  x402.yaml  serviceoffer-controller pin → :dev-d9527983fc74   (no @sha256: residue)
  llm.yaml   x402-buyer pin              → :dev-d9527983fc74
docker build -f Dockerfile.serviceoffer-controller -t …:dev-d9527983fc74 .  → ok
k3d image import …:dev-d9527983fc74 -c <cluster>                            → ok
(the exact build+import path buildAndImportLocalImages runs)
```

Release smoke: not re-run for this change (dev-build tagging only). The build/import
path it touches is the same one release-smoke exercises every run; the only behavioural
delta is the tag string, validated above. A full dev `obol stack up` is the recommended
final confirmation.

## Review Notes

Known gaps:
- `internal/defaults.devLocallyBuiltImageBases` and `internal/stack.baseLocalImages` remain
  duplicated (intentional, to avoid a defaults→stack import cycle) — unchanged by this PR.
- Uncommitted working-tree changes reuse the committed `dev-<sha>` (same caveat as the old
  `:latest`); `OBOL_FORCE_REBUILD_LOCAL_DEV_IMAGES` is the escape hatch.

Reviewer focus:
- `internal/defaults/defaults.go`: `DevImageTag` / `ReadDevImageTag`, the persist in
  `CopyInfrastructure`, and `rewriteDevDigestPins(defaultsDir, devTag)`.
- `internal/stack/stack.go`: `buildAndImportLocalImages` reads the persisted tag and builds/
  imports `<base>:<devTag>` (force-rebuild short-name matching still works via `localImageShortName`).

### DIFF
--- a/internal/defaults/defaults.go
+++ b/internal/defaults/defaults.go
@@ -25,7 +25,49 @@ const (
 	stackIDFile      = ".stack-id"
 	stackBackendFile = ".stack-backend"
 	stampFile        = ".obol-defaults-stamp"
+	// devImageTagFile records the tag the dev-mode manifest rewrite stamped
+	// the locally-built images with. internal/stack reads it at build time so
+	// the image it builds/imports matches what the rendered manifests pin —
+	// even if HEAD moved between `stack init` and `stack up`.
+	devImageTagFile = ".dev-image-tag"
 )
+
+// DevImageTag returns the tag used for locally-built dev images under
+// OBOL_DEVELOPMENT. It is `dev-<short-git-sha>` of the working tree, so each
+// branch/worktree builds a distinct tag and parallel dev stacks sharing one
+// Docker daemon never clobber each other's images (the `:latest` collision that
+// let a sibling worktree's build poison an unrelated stack). Committing changes
+// the SHA and triggers a fresh build; uncommitted changes reuse the committed
+// tag unless OBOL_FORCE_REBUILD_LOCAL_DEV_IMAGES is set. Falls back to `latest`
+// when the source is not a git checkout (e.g. a tarball build), preserving the
+// previous behaviour there.
+func DevImageTag() string {
+	cmd := exec.Command("git", "rev-parse", "--short=12", "HEAD")
+	out, err := cmd.Output()
+	if err != nil {
+		return "latest"
+	}
+	sha := strings.TrimSpace(string(out))
+	if !regexp.MustCompile(`^[0-9a-f]{7,40}$`).MatchString(sha) {
+		return "latest"
+	}
+	return "dev-" + sha
+}
+
+// ReadDevImageTag returns the dev image tag persisted at CopyInfrastructure
+// time, or "latest" if none was recorded (non-dev install, or pre-dates this
+// mechanism). internal/stack uses it to tag the images it builds.
+func ReadDevImageTag(cfg *config.Config) string {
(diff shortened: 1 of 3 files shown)


internal/defaults/defaults.go
```diff
@@ -63,16 +105,21 @@ func CopyInfrastructure(cfg *config.Config, backendName, stackID string) error {
...
-	// ghcr.io binaries. Rewrite digest pins to :latest after copy so the
-	// dev cycle Just Works without operators having to kubectl-set-image
-	// every loop.
+	// import them into k3d. The embedded templates pin published digests for
+	// production safety, which means the cluster ignores our locally-built
+	// images and silently uses stale ghcr.io binaries. Rewrite the digest
+	// pins to a per-commit `dev-<sha>` tag after copy so the dev cycle Just
+	// Works without operators having to kubectl-set-image every loop, and so
+	// parallel worktree stacks don't collide on a shared `:latest`. Persist
+	// the tag so internal/stack builds/imports the exact tag we pinned here.
```
OisinKyne, line 114: > Persist the tag so internal/stack builds/imports the exact tag we pinned here.

How does this get refreshed after persisting it once?

OisinKyne (approved):
One worry that i'm not so sure the write will overwrite an existing file, i guess it probably does?

PR #591, opened 2026-06-03
fix(dev): bake WalletConnect project ID into frontend dev builds

bussyjd:
## Summary
Redo of #579, scoped to the **local dev iteration** path. `just dev-frontend` / `dev-frontend-rebuild` build the local FE image but never passed `NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID`, so dev images shipped with the wallet UI disabled — even though the FE `Dockerfile` already consumes the arg (`ARG` → `ENV` → `pnpm run build`).

- Resolve the ID at build time, in order: `NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID` env → `WALLET_CONNECT_PROJECT_ID` env → frontend repo `.env.local`.
- Pass it as a `--build-arg`. The ID is public by design (browsers use it to reach WalletConnect) but stays out of tracked obol-stack files.
- Factor the two near-identical recipes into a shared private recipe `_dev-frontend-build no_cache set_image` (vs. #579 which duplicated the resolver into both recipes).

## Scope
Local dev iteration only. The production release image is injected separately in the frontend repo's own build (per the discussion on #579).

## Validation
- `just --summary` / `--list` parse cleanly; `_dev-frontend-build` stays private.
- `just --dry-run dev-frontend` → cache + set-image; `dev-frontend-rebuild` → `--no-cache` + skip-set-image.
- Resolver extracts the real 32-char ID from `.env.local`.
- Real `docker build --build-arg …` against the FE repo → confirmed the project ID is baked into the compiled `.next/static` bundle.

### DIFF
--- a/justfile
+++ b/justfile
@@ -43,29 +43,71 @@ frontend_dir := env("FRONTEND_DIR", justfile_directory() / "../obol-stack-front-
 dev_image    := "localhost:54103/obol-stack-front-end:dev"
 
 # Build frontend from local source, push to local registry, and restart the pod
-dev-frontend:
-    #!/usr/bin/env bash
-    set -e
-    echo "→ Building {{ dev_image }} from {{ frontend_dir }}"
-    docker build -t {{ dev_image }} {{ frontend_dir }}
-    echo "→ Pushing {{ dev_image }} to local registry"
-    docker push {{ dev_image }}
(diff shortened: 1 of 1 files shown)


OisinKyne (approved):
I wasn't thinking about the local dev front end flow when i closed this earlier, my b.

PR #624, opened 2026-06-11
feat: import/export persistence clean up

OisinKyne:
## Problem to be solved

Losing agent data on a purge. And other resources giving up on the persistence aspects.

This fable run was intended to add import / export, and phase 2 was to bring more persistence back, to make the cluster more declarative than imperative. Next step is to rebase (ish) off the other pr in flight #623 

## Summary

Phase 2 is done and staged on
  feat/stack-export-import (on top of your
  Phase 1 commit ba9b536) — stopping
  short of committing per your yubikey
  workflow.

  What record-on-write now covers

  Agent CRs — obol agent new/update
  persist the applied manifest to
  $CONFIG_DIR/agents/<name>.yaml; agent 
  delete removes it; agentcrd.ResumeAll
  replays them at stack up. This was the
  worst gap: host seeds existed but
  nothing recreated the CR, so agents (and
  any offer pointing at them) silently
  never returned.

  sell agent offers — now persist into the
  existing sell-http/ store, so the
  proven replay path and sell delete
  cleanup handle them with zero new
  machinery. sell mcp (foreground process)
  and sell demo (ephemeral) stay
  unrecorded by design.

  Model config — obol model
  setup|prefer|remove snapshot the
  resulting model_list plus referenced
  provider API keys to
  $CONFIG_DIR/llm/recorded-models.yaml
  (0600). paid/* entries are filtered out
  — verified live against your cluster,
  where the real paid/AEON-7/... purchase
  route was correctly excluded while the
  Ollama entry was captured. The subtle
  design point: recording happens only at
  the command layer, never from
  autoConfigureLLM — otherwise stack-up
  auto-detection would overwrite your
  preferences moments before
  model.ReconcileRecorded (which runs
  after auto-config, before default-Hermes
  setup, so your recorded head model
  stays the agents' default).

  Remote RPCs — AddPublicRPCs/AddCustomRPC
  /RemovePublicRPCs maintain
  $CONFIG_DIR/rpc/recorded-upstreams.yaml
  (0600, since custom URLs can carry
  paid-provider keys), with ChainList
  endpoints snapshotted at add-time for
  deterministic offline replay. stack up
  replays through the same idempotent add
  functions. Local-node upstreams stay
  unrecorded — network sync already
  re-registers those.

  Stack-up ordering is pinned by a
  source-guard test
  (cmd/obol/stackup_resume_guard_test.go,
  mirroring your existing resumeSellOffers
  guard): stack.Up → RPCs → Agent CRs →
  sell offers, with Agent CRs strictly
  before offers since agent-backed
  ServiceOffers resolve agent.ref.

  Validation

  All 35 packages pass. Live smoke on your
  dev cluster: model record verified with
  the paid-filter; agent full cycle
  verified (agent new → record written →
  CR deleted from etcd → replayed from
  record → agent delete → record and data
  dir cleaned). Cluster left healthy
  (hermes 2/2 Running). stack export needs
  no changes — the config component picks
  up all three new stores automatically.

  One noted follow-up in the plan doc:
  (agent new → record written → CR deleted from etcd → replayed
  from record → agent delete → record and data dir cleaned).
  Cluster left healthy (hermes 2/2 Running). stack export needs
  no changes — the config component picks up all three new stores
  automatically.

  One noted follow-up in the plan doc: custom RPC endpoints have
  no removal command (network remove only drops ChainList
  selections, matching its cluster-side semantics), so a recorded
  custom endpoint currently requires editing the file — a
  network remove --endpoint flag would close that. Phase 3
  (sellable OCI bundles + obol sell export) remains the open
  phase.


bussyjd (approved):
## Approve — reviewed in depth, hardened, and coverage-filled

Read the full net-new persistence subsystem (stackbackup / agentcrd / model.record / network.record + the stack-up wiring). The design is sound — record-on-write with atomic 0600 writes, the agent-CRs-before-offers and model-reconcile-after-autoconfig ordering invariants, honest secret disclosure, and version gates on all three record formats. Test quality on the pure-logic layer is genuinely strong (real contracts asserted: 0600 perms, version rejection, merge ordering/dedup, unrecord-keeps-custom), not coverage padding.

Two things were addressed during review (pushed to this branch, both GitHub-verified):

- **`71b5848` — security:** the import extractor sanitized entry *names* but wrote symlinks from the raw `Linkname`, allowing a symlink-target escape (clean-named link → `/etc`, then a write *through* it = arbitrary file write as the importing user). Fixed with `symlinkEscapesRoot` (in-root links still work) + a decompression-ratio guard against zip-bombs. Tests added; `TestArchiveRoundTrip` unchanged.
- **`975aefd` — coverage:** unit tests for the previously-0% pure helpers (`Manifest.component`, `listInstances`, non-gzip rejection, `walletNote`, `mapKeys`) and `flow-18-stack-export-import.sh` — the export→purge→import→up→verify round-trip the unit tests can't reach. The flow is standalone/destructive (not in the release-smoke baseline) and needs one live-cluster validation run before use as a gate.

Remaining items are design judgments, not blockers, left to you: RPC record fail-loud vs. model/agent best-effort; the cluster-dump-vs-`recorded-models.yaml` double path for LiteLLM config. Worth a rebase onto main once the rc15 train (#623) lands so the diff collapses to just the three persistence commits.

LGTM.

PR #637, opened 2026-06-12
feat(stack): add `obol node` for k3s multi-node clusters

bussyjd:
Adds a first-class CLI for adding worker nodes to a **k3s-backed** stack — the multi-node "master runs pods on another node" path that's fatal on the k3d/Docker-Desktop master but works natively over the LAN.

## What
- **`obol node token`** — prints the k3s agent-join one-liner (`K3S_URL` + `K3S_TOKEN`, version-pinned to the running server). `--json` for machine use, `--server-url` override.
- **`obol node list`** — nodes with their `obol.tech/accelerator` label.
- Both **guard on the backend**: on k3d they fail fast (`a Docker-Desktop master cannot accept remote node joins — its flannel overlay is not routable off-host`).
- **tls-san plumbing**: `k3s-config.yaml` now templates `{{NODE_IP}}`/`{{NODE_HOSTNAME}}` (substituted in `backend_k3s.Init`) so workers can join by LAN IP or hostname. (k3s already auto-adds the primary node IP, so this mainly adds an explicit hostname SAN + cert determinism.)

## Why
On Docker Desktop the k3d server's flannel VTEP is a VM-internal `172.x` IP, unroutable from the LAN — a joined agent registers Ready but cross-node pod networking hangs. On a **native k3s master** it all works. This makes that path ergonomic.

## Tests / validation
- 7 new table-driven tests (`internal/stack/node_test.go`, `backend_k3s_init_test.go`); `go build ./...` + full `internal/stack` suite green; backend guard verified live.
- **Validated end-to-end** on two LAN boxes: native k3s master scheduling cross-node pods (CoreDNS + flannel VXLAN) and a 4-bit QLoRA fine-tune onto a remote RTX 2060, adapter persisted to a node-local local-path PVC.

Faithful to the existing `Backend` interface and `func xCommand(cfg) *cli.Command` pattern.


LuuOW (commented):
Technical audit: code implementation verified for system consistency.

